PR #14, opened 2024-11-04
Documentation and minor robustness changes

GottfriedHerold:
This PR mostly adds documentation changes that I wrote while trying to understand the codebase.
It intentionally does not do functional changes.

The only actual code changes are additions of some assert! guards, adding some extra tests
and splitting some functions more cleanly into parallel / non-parallel versions.
The latter is done to simplify testing and enable differential tests (which the old version could not do, because only one version was compiled).

### DIFF
--- a/src/poly_utils/hypercube.rs
+++ b/src/poly_utils/hypercube.rs
@@ -10,7 +10,7 @@
 /// The point is encoded via the bits of a usize and we do not store n.
 pub struct BinaryHypercubePoint(pub usize);
 
-// BinaryHypercube is an Iterator that is used to range over the points of {0,1}^n, where n == `num_variables`
+/// BinaryHypercube is an Iterator that is used to range over the points of {0,1}^n, where n == `num_variables`
 pub struct BinaryHypercube {
     pos: usize,  // current position, encoded via the bits of pos
     num_variables: usize, // dimension of the hypercube
--- a/src/poly_utils/hypercube.rs
+++ b/src/poly_utils/hypercube.rs
@@ -3,14 +3,14 @@
 // TODO (Gotti): Should pos rather be a u64? usize is platform-dependent, giving a platform-dependent limit on the number of variables.
 // num_variables may be smaller as well.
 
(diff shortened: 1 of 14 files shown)


src/poly_utils/hypercube.rs
```diff
@@ -1,13 +1,24 @@
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+
+// TODO (Gotti): Should pos rather be a u64? usize is platform-dependent, giving a platform-dependent limit on the number of variables.
```
WizardOfMenlo, line 3: Yes, I would guess that even u32 would be enough, but let's do u64 to be safe.

src/utils.rs
```diff
@@ -16,22 +21,37 @@ pub fn to_binary(value: usize, n_bits: usize) -> Vec<bool> {
     result
 }
 
+// TODO(Gotti): n_bits is a misnomer if base > 2. Should be n_limbs or sth.
+// Also, should the behaviour for value >= base^n_bits be specified as part of the API or asserted not to happen?
```
WizardOfMenlo, line 25: Let's assert it out of existence ;)

src/utils.rs
```diff
@@ -16,22 +21,37 @@ pub fn to_binary(value: usize, n_bits: usize) -> Vec<bool> {
...
         result[n_bits - 1 - i] = (value % (base as usize)) as u8;
         value /= base as usize;
     }
+    // TODO: Should we assert!(value == 0) here to check that the orginally passed `value` is < base^n_bits ?
 
     result
 }
 
+// Gotti: Consider renaming this function. The name sounds like it's a PRG.
+// TODO (Gotti): Check that ordering is actually correct at point of use (everything else is big-endian).
```
WizardOfMenlo, line 52: This randomness can actually be in whatever order we want, as long as it is self-consistent

WizardOfMenlo (approved):
LGTM!

PR #16, opened 2024-11-25
Merge PCS and LDT

recmo:


recmo:
In it's current state this PR adds an `initial_statement` flag to the parameters and replaces the LDT prover with the regular prover, while keeping the LDT verifier. It passes for `--field Field256` but fails for Goldilocks.

WizardOfMenlo (approved):
LGTM!

PR #17, opened 2024-11-25
Use FS to generate STIR queries

kustosz:
Migrates away from using ChaCha20 to generating STIR queries using the FS transcript. This is a huge boon for recursive verification. The query distribution is unbiased for power-of-two sized domains.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ outputs/temp/
 *.pdf
 scripts/__pycache__/
 .DS_Store
+.idea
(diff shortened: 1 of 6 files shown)


WizardOfMenlo:
LGTM! Will point out that for recursive verification there are also a couple of annoying things that we might want to handle:
- The path pruning makes the Merkle paths to verify hard to do in circuit (can be fixed by decompressing the Merkle path or Merkle capping)
- The dedup will make the number of queries also vary, which we should be able to handle similarly.

WizardOfMenlo (approved):

PR #22, opened 2024-11-27
Generic type bounds for FS types

kustosz:
Changes the type bounds on `prove` and `verify` to not hardcode keccak for challenges. This is required to allow these methods to work with algebraic hashes.

### DIFF
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -5,7 +5,7 @@ pub trait OODIOPattern<F: Field> {
     fn add_ood(self, num_samples: usize) -> Self;
 }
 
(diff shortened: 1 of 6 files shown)


src/skyscraper.rs
WizardOfMenlo, line 1: I'm guessing this is not meant for this PR?

WizardOfMenlo (approved):
LGTM! Let's just remove the skyscraper.rs file

src/skyscraper.rs
kustosz: yes this is accidental LMAO

kustosz: Removed :)

PR #24, opened 2024-11-29
Generalize the type of merkle roots

kustosz:
Another (final, I think!) step towards algebraic hash support – it lifts the requirement of merkle roots to be bytes, making them fully configurable by the clients.


WizardOfMenlo (approved):
LGTM

PR #29, opened 2025-02-04
Add support for weighted queries

veljkovranic:
The current code version supports only f(x) = y statements.

The goal of the PR is to add support for 
```math
$$ \sum_{b \in \{0, 1\}^n} w(f(b), b) = 0 $$
```


src/poly_utils/evals.rs
```diff
@@ -46,6 +46,78 @@ where
...
     }
+    
+    fn eval_multilinear(&self, evals: &[F], point: &[F]) -> F {
+        debug_assert_eq!(evals.len(), 1 << point.len());
+        let one = F::one();
+        match point {
+            [] => evals[0],
+            [x] => evals[0] * (one - *x) + evals[1] * *x,
+            [x0, x1] => {
+                let a0 = evals[0] * (one - *x1) + evals[1] * *x1;
```
recmo, line 57: We can save a mul:

```rust
let a0 = evals[0] + (evals[1] - evals[0]) * *x1;
```

WizardOfMenlo (commented):
API comments:
- I would like the statement and statement verifier to be hard to misuse. Ideally, you could only create a statement verifier from a statement struct (and it will be filled in with the right fields)
- I would move the statement_values_at_random point to the transcript
- I think we should combine SumcheckSingle::new and add_new_equality into a single method, we never use them distinctly
- statement stuff should deduplicate with the things in poly_utils/mod.rs

WizardOfMenlo (approved):
LGTM! Happy with it now. I'll open an issue (#61) about adding the evaluations of the weights at the extensions in the transcript, but we can merge for now

Issue #38, opened 2025-03-09
Identify performance regression


WizardOfMenlo:
We should setup some more mature GitHub actions to identify performance regressions

PR #41, opened 2025-03-10
optimize `transpose_square_swap_parallel`

tcoratger:
@WizardOfMenlo I think we can delete the benchmark (this was just to ensure myself that I'm optimizing the function) because if we want to keep it we have to keep some stuffs public which maybe makes no sense.

Here is the results I obtain between the old version vs the new version

```shell 
  Running benches/transpose_square_swap.rs (target/release/deps/transpose_square_swap-0659aeb663ba5a23)
Gnuplot not found, using plotters backend
transpose_square_swap_parallel 1024x1024
                        time:   [987.70 µs 1.0109 ms 1.0370 ms]
                        change: [-22.682% -20.254% -17.833%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe
```


WizardOfMenlo (approved):
Looks good to me! Maybe, we could remove the matrix swap benchmark, and instead add a NTT benchmark (testing the NTT end to end). This would also go as an initial step in #38

tcoratger:
> Looks good to me! Maybe, we could remove the matrix swap benchmark, and instead add a NTT benchmark (testing the NTT end to end). This would also go as an initial step in #38

@WizardOfMenlo I've just removed the benchmark, will work a bit later on some global benchmarks about ntt, I've few other ideas in mind where we can optimize first, feel free to merge this one if you feel ok with the improvements and will work on some follow ups.

PR #43, opened 2025-03-10
fix: remove const qualifier from ptr_at_mut

yugocabrio:
### Changes
- Removes `const` qualifier from `MatrixMut::ptr_at_mut`

### Why this is needed
Fixes E0658 compile error: mutable references in const functions are unstable


WizardOfMenlo:
Which rust version are you using? I compile successfully on 1.83 stable and 1.85 nightly.

PR #45, opened 2025-03-10
fix `sqrt_factor` and add fuzzing

tcoratger:


recmo (approved):
This is great! Thanks for fixing this!

I was lazy in my handling of sqrt(3), mostly because:

* I implemented radix-3 as a gimmick, not expecting anyone to use non-power-of two radices (though they have their uses).
* The NTT will succeed as long as the function returns a nontrivial factor. Having it be the largest <= sqrt(n) just improves performance.

Having it fixed is cool! And hopefully makes non-power-of-two NTTs more useable.

Also appreciate the use of proptest, superb way to test functions like this 👍.

PR #46, opened 2025-03-10
Optimize Boolean Hypercube Summation

yugocabrio:
## Overview
This PR optimizes the summation over the Boolean hypercube (\{0,1\}^n) by adding a more efficient method that avoids enumerating all ternary points (\{0,1,2\}^n) and filtering them afterwards.

## Changes
- Added a new method `sum_over_boolean_hypercube()` that directly iterates through 2^n indices.
- Uses a big-endian bit reading to stay consistent with the existing ternary decomposition.

## Performance Improvement
[A quick benchmark](https://github.com/yugocabrio/whir/blob/0397a3cddbba9e257b3c46c6102053eb51f1ad46/src/sumcheck/proof.rs#L134C1-L160C6) (run in a loop of 10 iterations) shows speedups:

- Original implementation(`sum_over_hypercube`): ~1.86 ms  
- Optimized implementation(`sum_over_boolean_hypercube()`): ~0.08 ms

### DIFF
--- a/src/sumcheck/mod.rs
+++ b/src/sumcheck/mod.rs
@@ -28,7 +28,7 @@ mod tests {
         let poly_1 = prover.compute_sumcheck_polynomial(folding_factor);
 
         // First, check that is sums to the right value over the hypercube
-        assert_eq!(poly_1.sum_over_hypercube(), claimed_value);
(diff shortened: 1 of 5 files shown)


WizardOfMenlo (approved):
LGTM! Thanks!

PR #54, opened 2025-03-11
`LagrangePolynomialIterator`: revisit documentation and add tests

tcoratger:
- `From` is used instead of `new` for easier manipulations
- Doc is revisited to format the doc comment properly
- Unit tests are added to cover all possible cases

### DIFF
--- a/src/poly_utils/sequential_lag_poly.rs
+++ b/src/poly_utils/sequential_lag_poly.rs
@@ -7,7 +7,6 @@ use super::{hypercube::BinaryHypercubePoint, multilinear::MultilinearPoint};
 /// Iterator for evaluating the Lagrange polynomial over the hypercube `{0,1}^n`.
 ///
 /// This efficiently computes values of the equality polynomial at every binary point.
-/// Instead of iterating in numerical order, it follows **Gray code ordering** to minimize updates.
 ///
 /// Given a multilinear point `(c_1, ..., c_n)`, it iterates over all binary vectors `(x_1, ...,
 /// x_n)` and computes:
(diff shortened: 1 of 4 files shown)


src/poly_utils/sequential_lag_poly.rs
```diff
@@ -4,47 +4,67 @@ use ark_ff::Field;
...
 use super::{hypercube::BinaryHypercubePoint, multilinear::MultilinearPoint};
 
-/// There is an alternative (possibly more efficient) implementation that iterates over the x in Gray code ordering.
+/// Iterator for evaluating the Lagrange polynomial over the hypercube `{0,1}^n`.
 ///
-/// LagrangePolynomialIterator for a given multilinear n-dimensional `point` iterates over pairs (x, y)
-/// where x ranges over all possible {0,1}^n
-/// and y equals the product y_1 * ... * y_n where
+/// This efficiently computes values of the equality polynomial at every binary point.
+/// Instead of iterating in numerical order, it follows **Gray code ordering** to minimize updates.
```
WizardOfMenlo, line 10: This implementation is the binary ordering one, not the Gray code ordering right?

WizardOfMenlo (commented):
I think we have to update the docs, this implementation does not do Gray code iteration afaicr

src/poly_utils/sequential_lag_poly.rs
tcoratger: Yep true it seemed weird to me, just fixed

WizardOfMenlo (approved):

PR #57, opened 2025-03-11
poly: simplify `compute_fold` and better doc

tcoratger:


src/poly_utils/fold.rs
```diff
@@ -17,28 +38,32 @@ pub fn compute_fold<F: Field>(
 ) -> F {
     let mut answers = answers.to_vec();
 
-    // We recursively compute the fold, rec is where it is
+    // Perform the folding process `folding_factor` times.
     for rec in 0..folding_factor {
         let offset = answers.len() / 2;
-        let mut new_answers = vec![F::ZERO; offset];
```
tcoratger, line 23: I think this new allocation was useless and so it is removed

WizardOfMenlo (approved):

PR #65, opened 2025-03-12
sumcheck: some refactorization and touchups

tcoratger:
- Functions that don't use `self` internally are moved outside of struct implementation for clarity
- `add_new_equality` is refactorized
- small touchups

### DIFF
--- a/src/sumcheck/sumcheck_single.rs
+++ b/src/sumcheck/sumcheck_single.rs
@@ -79,15 +79,6 @@ where
         SumcheckPolynomial::new(vec![eval_0, eval_1, eval_2], 1)
     }
 
-    /// Computes the sumcheck polynomial `S(X)`, which is quadratic.
-    ///
-    /// The sumcheck polynomial is given by:
-    ///
-    /// ```text
-    /// S(X) = ∑ p(β) * eq(β)
-    /// ```
-    ///
-    /// where `β` are points in `{0,1,2}^1`, and `S(X)` is a quadratic polynomial.
     #[cfg(feature = "parallel")]
     pub fn compute_sumcheck_polynomial(&self) -> SumcheckPolynomial<F> {
         assert!(self.num_variables() >= 1);
@@ -363,8 +354,7 @@ mod tests {
         let polynomial =
             CoefficientList::new(vec![F::from(1), F::from(5), F::from(10), F::from(14)]);
 
-        let claimed_value: ark_ff::Fp<ark_ff::MontBackend<crate::crypto::fields::FConfig64, 1>, 1> =
-            polynomial.evaluate(&eval_point);
+        let claimed_value = polynomial.evaluate(&eval_point);
 
         let eval = polynomial.evaluate(&eval_point);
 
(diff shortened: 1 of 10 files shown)


src/sumcheck/sumcheck_single.rs
```diff
@@ -137,66 +144,33 @@ where
...
-            self.sum += *rand * eval;
-        }
+
+        points
+            .iter()
+            .zip(combination_randomness.iter().zip(evaluations))
+            .for_each(|(point, (&rand, &eval))| {
+                eval_eq(&point.0, self.weights.evals_mut(), rand);
+                self.sum += rand * eval;
+            });
```
tcoratger, line 173:
```suggestion
for (point, rand) in points.iter().zip(combination_randomness) {
            // TODO: We might want to do all points simultaneously so we
            // do only a single pass over the data.
            Self::eval_eq(&point.0, self.weights.evals_mut(), *rand);
        }
        // Update the sum
        for (rand, eval) in combination_randomness.iter().zip(evaluations.iter()) {
            self.sum += *rand * eval;
        }
```

src/sumcheck/sumcheck_single.rs
```diff
@@ -72,7 +79,15 @@ where
         SumcheckPolynomial::new(vec![eval_0, eval_1, eval_2], 1)
     }
 
-    /// Compute the polynomial that represents the sum in the first variable.
+    /// Computes the sumcheck polynomial `S(X)`, which is quadratic.
+    ///
+    /// The sumcheck polynomial is given by:
+    ///
+    /// ```text
+    /// S(X) = ∑ p(β) * eq(β)
```
WizardOfMenlo, line 87: Note, here we are not doing eq anymore, we have generic weights. Also, add X to rhs

src/sumcheck/sumcheck_single.rs
```diff
@@ -318,7 +363,8 @@ mod tests {
         let polynomial =
             CoefficientList::new(vec![F::from(1), F::from(5), F::from(10), F::from(14)]);
 
-        let claimed_value: ark_ff::Fp<ark_ff::MontBackend<crate::crypto::fields::FConfig64, 1>, 1> = polynomial.evaluate(&eval_point);
+        let claimed_value: ark_ff::Fp<ark_ff::MontBackend<crate::crypto::fields::FConfig64, 1>, 1> =
```
WizardOfMenlo, line 366: Shouldn't this just be of type F?

src/sumcheck/sumcheck_single.rs
```diff
@@ -349,7 +395,7 @@ mod tests {
     fn test_eval_eq() {
         let eval = vec![F::from(3), F::from(5)];
         let mut out = vec![F::ZERO; 4];
-        SumcheckSingle::eval_eq(&eval, &mut out, F::ONE);
+        eval_eq(&eval, &mut out, F::ONE);
         dbg!(&out);
```
WizardOfMenlo, line 399: Remove dbg in test?

WizardOfMenlo (changes requested):
Few small changes but LGTM

src/sumcheck/sumcheck_single.rs
tcoratger: fixed

src/sumcheck/sumcheck_single.rs
tcoratger: fixed

WizardOfMenlo (approved):

PR #66, opened 2025-03-12
github ci: add clippy and fmt checks

tcoratger:


.github/workflows/rust.yml
```diff
@@ -2,23 +2,54 @@ name: Rust
...
-    - name: Switch toolchain
-      run: rustup update nightly && rustup default nightly
-    - name: Build
-      run: cargo build --release --verbose
-    - name: Run tests
-      run: cargo test --release --verbose
+      - name: Fetch Repository
+        uses: actions/checkout@v4
+
+      - name: Install stable toolchain
```
WizardOfMenlo, line 33: We should do nightly, maybe also update the build and test actionis to use the actions-rust-lang?

.github/workflows/rust.yml
```diff
@@ -2,23 +2,54 @@ name: Rust
...
+
+  cargo-fmt:
+    name: Cargo fmt
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Fetch Repository
+        uses: actions/checkout@v4
+
+      - name: Install stable toolchain
```
WizardOfMenlo, line 49: Same

WizardOfMenlo (commented):
Let's move to nightly and fix the errors and then we are good

.github/workflows/rust.yml
tcoratger: fixed

.github/workflows/rust.yml
tcoratger: fixed

PR #69, opened 2025-03-13
prover core: reactivate

tcoratger:
- reactivation of the prover core module
- @WizardOfMenlo for `prover_not_skipping` module, imo if we are sure we will never use it again, we should rm it because right now it is not up to date with `SumcheckSingle`. So we're going to lose track over time, and it seems the tests need to be reworked at least a little bit. So, in my opinion either we reactivate it now in a follow-up, but we need more work on the tests or we delete it permanently, right?


WizardOfMenlo:
I am in favour of removing all the sumcheck variants that we don't use. We can always copy the code back later on if needed. The variant skipping rounds end up very expensive on the prover side (and has larger proofs), the only place it might be worthwhile is in the first round (since it can avoid working in extensions for a bit), but I don't think this is worth it for now.

WizardOfMenlo:
The prover core also seems superseded, so I propose to remove it as well.

tcoratger:
@WizardOfMenlo Fine with me, just removed all the unused variants

PR #79, opened 2025-03-15
ntt expand from coeff: add criterion bench

tcoratger:
Path to https://github.com/WizardOfMenlo/whir/issues/38

@WizardOfMenlo Just added this criterion benchmark to track the ntt perfs over a wide range of parameters. This should help us in the future to better test our optimization strategies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,10 +15,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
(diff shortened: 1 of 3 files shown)


WizardOfMenlo:
Let's scale it up, ideally 2^20 at least

tcoratger:
> Let's scale it up, ideally 2^20 at least

@WizardOfMenlo Fixed

PR #87, opened 2025-03-17
`evaluate_at_extension`: simplify usage and add unit tests

tcoratger:


src/poly_utils/coeffs.rs
```diff
@@ -325,9 +310,11 @@ mod tests {
             multilinear::MultilinearPoint,
         },
     };
+    use ark_ff::AdditiveGroup;
     use ark_poly::{univariate::DensePolynomial, Polynomial};
 
     type F = Field64;
+    type E = Field64;
```
WizardOfMenlo, line 317: Maybe we should this be a quadratic extension of F? Just to test the intended use case?

tcoratger: yep, fixed

WizardOfMenlo (approved):

PR #88, opened 2025-03-17
committer: better doc and unit test

tcoratger:


src/whir/committer.rs
```diff
@@ -114,3 +148,193 @@ where
...
+
+        let committer = Committer::new(params);
+        let witness = committer.commit(&mut merlin, polynomial).unwrap();
+
+        // Expansion factor is 2
+        assert_eq!(
+            witness.merkle_leaves.len(),
+            1024 * 2,
+            "Merkle tree should have expected number of leaves"
+        );
```
WizardOfMenlo, line 299: Can we refactor the tests in `basic_commitment` into a `validate_commitment` method, which then we apply to each of these instances?

tcoratger: Mhhh I feel like it'll be confusing if I do this because:
- in `test_large_polynomial` I'm testing specifically and only the Merkle leaves length to check the expansion factor.
- in `test_commitment_without_ood_samples` I'm testing specifically and only the fact that `ood_points` is empty.

So it seems better to isolate the cases, right? What do you think?

WizardOfMenlo (approved):

PR #90, opened 2025-03-17
whir params: refactor and unit tests

tcoratger:
- Some functions are a bit simplified
- Some unit tests are added (we don't have a full coverage yet)
- Some functions are made const when possible

@WizardOfMenlo I think this part will need to be reworked in the future. I wanted to add documentation to transcribe the usefulness of each function and parameter, but for now it remains a little unclear to me. I have the impression that things will need to be split up with probably references to the paper to understand where all the calculations on floating numbers come from. I haven't had time to look at it yet, we can take care of it later.


WizardOfMenlo:
Hi, yes. Maybe we should pull off the work on parameters a bit. I am doing a full rewrite of this section (adding more parameters and a more sane security estimate). We should do a pass once that is merged.

WizardOfMenlo (approved):

PR #95, opened 2025-03-18
sumcheck single: merge `compress` implementations

tcoratger:


src/sumcheck/sumcheck_single.rs
```diff
@@ -202,38 +202,6 @@ where
...
-            .collect();
-        let evaluations_of_eq = self
-            .weights
-            .evals()
-            .chunks_exact(2)
-            .map(|at| (at[1] - at[0]) * randomness + at[0])
-            .collect();
-
-        // Update
-        self.num_variables -= 1;
```
tcoratger, line 231: @WizardOfMenlo I think this line was deprecated after recent modifications. I have the impression that currently the tests do not run on both cfg in the CI (parallel and non-parallel). In a follow-up PR it will probably be necessary to fix this to have the runs in both cases and catch errors of this type because I think that in non-parallel cfg, the code did not compile.

WizardOfMenlo (approved):

PR #96, opened 2025-03-18
mv `ParsedProof` to dedicated file with tests

tcoratger:
Verifier file is already quite big so that `ParsedProof` is moved to a dedicated file with unit tests

### DIFF
--- a/src/whir/parsed_proof.rs
+++ b/src/whir/parsed_proof.rs
@@ -1,0 +1,221 @@
+use crate::{
+    poly_utils::{coeffs::CoefficientList, multilinear::MultilinearPoint},
+    sumcheck::SumcheckPolynomial,
+};
+use ark_ff::FftField;
+
+#[derive(Default, Debug, Clone)]
+pub(crate) struct ParsedRound<F> {
+    pub(crate) folding_randomness: MultilinearPoint<F>,
+    pub(crate) ood_points: Vec<F>,
+    pub(crate) ood_answers: Vec<F>,
+    pub(crate) stir_challenges_indexes: Vec<usize>,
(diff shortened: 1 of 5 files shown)


src/whir/parsed_proof.rs
```diff
@@ -0,0 +1,221 @@
...
+                    .iter()
+                    .map(|answers| {
+                        CoefficientList::new(answers.clone()).evaluate(&round.folding_randomness)
+                    })
+                    .collect()
+            })
+            .collect();
+
+        // Add final round if needed
+        if !self.final_randomness_answers.is_empty() {
```
tcoratger, line 53: @WizardOfMenlo Here please double check, but it seems that if `final_randomness_answers` we don't need to push anything. Before even if `final_randomness_answers` was empty, we had at the end `[[]]`

WizardOfMenlo: Why can final_randomness_answers be empty? In theory these are the last Merkle answers which should always be there?

tcoratger: Yes, in theory, in the protocol, I imagine this should never happen, but I came across this edge case in the unit tests, so I thought it was better to do it this way.

But let me know if you feel this is useless, and I can remove this check.

WizardOfMenlo: I think we should just disallow this in the parsing, the protocol doesn't make much sense if the final answers (or any really are empty), and the verifier should reject.

tcoratger: yeah that makes sense, for now, just reverted this to keep as it was before for this check

WizardOfMenlo (approved):

PR #111, opened 2025-03-21
compute_stir_evaluations: separate stir eval from prover core

tcoratger:
Another step toward https://github.com/WizardOfMenlo/whir/pull/97


WizardOfMenlo:
I wonder if it makes sense to refactor this to a common method of prover and verifier. I think at the time I just copy and pasted it from the verifier

tcoratger:
> I wonder if it makes sense to refactor this to a common method of prover and verifier. I think at the time I just copy and pasted it from the verifier

@WizardOfMenlo Mhhh that would be great yes but you mean this verifier method? https://github.com/WizardOfMenlo/whir/blob/d2424c545bed5ac274d2508d99ea8eb945a77a9b/src/whir/verifier.rs#L372

It seems they are different no?

WizardOfMenlo:
They are a bit different now, but the core is the same. At each round the prover needs to compute the evaluations, which it does by reading the corresponding Merkle leaves, and then either computing the fold either as in ProverHelps or as in Naive. 
The verifier does the same (but in the compute_folds we do all of them at once because we already have all the information). 
Ah also, tiny opt, we can take the F::from(2).inverse().unwrap() out of the loop (ideally precompute just once) to save some cycles.

WizardOfMenlo (approved):

PR #123, opened 2025-03-22
prover: precompute folding factor

tcoratger:


WizardOfMenlo (approved):
Good, let's fix the merge conflicts and then good to merge

tcoratger:
> Good, let's fix the merge conflicts and then good to merge

@WizardOfMenlo fixed

PR #125, opened 2025-03-23
prover: avoid unecessary randomness zero init

tcoratger:
It is useless to init first all `randomness_vec` with zeroes because there are directly overwritten after that, this is avoided here.


WizardOfMenlo (approved):
Hugh, not sure why I was doing that...

PR #148, opened 2025-03-28
Add support for `FoldType::ProverHelps` optimization

jdodinh:
I added the fold optimization for the first round of the prover (plus some refactoring). 

This means that if the `FoldType` is set to `ProverHelps`, then the prover interpolates over the coset, as opposed to sending the evaluations of the polynomial. This alleviates the work of the verifier, who can now simply evaluate the polynomial at the random point.

Feedback is appreciated on whether the Verifier needs to perform some extra check in this scenario.

### DIFF
--- a/src/stir_ldt/mod.rs
+++ b/src/stir_ldt/mod.rs
@@ -1,5 +1,6 @@
 use ark_crypto_primitives::merkle_tree::{Config, MultiPath};
 use ark_ff::Field;
+use ark_poly::univariate::DensePolynomial;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 
 pub mod committer;
@@ -16,6 +17,7 @@ where
     MerkleConfig: Config<Leaf = [F]>,
 {
     merkle_proofs: Vec<(MultiPath<MerkleConfig>, Vec<Vec<F>>)>,
+    first_round_coeffs: Option<Vec<DensePolynomial<F>>>,
 }
 
 pub fn stir_proof_size<MerkleConfig, F>(
--- a/src/stir_ldt/mod.rs
+++ b/src/stir_ldt/mod.rs
@@ -1,6 +1,5 @@
 use ark_crypto_primitives::merkle_tree::{Config, MultiPath};
 use ark_ff::Field;
-use ark_poly::univariate::DensePolynomial;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 
 pub mod committer;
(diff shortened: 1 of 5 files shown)


src/stir_ldt/mod.rs
```diff
@@ -16,6 +17,7 @@ where
     MerkleConfig: Config<Leaf = [F]>,
 {
     merkle_proofs: Vec<(MultiPath<MerkleConfig>, Vec<Vec<F>>)>,
+    first_round_coeffs: Option<Vec<DensePolynomial<F>>>,
```
WizardOfMenlo, line 20: Why does this need a addition to the proof? In the WHIR impl we simply do it within the (already existing) Merkle paths, we shouldn't need it.

src/stir_ldt/prover.rs
```diff
@@ -83,12 +58,14 @@ where
         }
 
         let mut ctx = RoundContext {
+            round_1: true,
```
WizardOfMenlo, line 61: Can we just move the first round out of the loop? Seems cleaner

src/stir_ldt/prover.rs
```diff
@@ -83,12 +58,14 @@ where
...
             f_domain: self.0.starting_domain.clone(),
             r_fold,
             f_poly: witness.polynomial.clone(),
             merkle: witness.merkle_tree.clone(),
             evals: witness.merkle_leaves.clone(),
             merkle_proofs: vec![],
+            first_round_coeffs: None,
```
WizardOfMenlo, line 68: This probably disappears

src/stir_ldt/prover.rs
```diff
@@ -288,9 +209,128 @@ where
...
+        let g_evals_folded = utils::stack_evaluations(g_evals, self.0.folding_factor);
+
+        // TODO: if fold_type is always Naive, this is just an assertion, nothing happens here
+        //
+        // At this point folded evals is a matrix of size (new_domain.size()) X (1 << folding_factor)
+        // This allows for the evaluation of the virutal function using an interpolation on the rows.
+        // TODO: for stir we do only Naive, so this will need to be adapted.
+        let g_evals_folded = restructure_evaluations(
+            g_evals_folded,
+            FoldType::Naive,
```
WizardOfMenlo, line 240: Why don't we pass here the right fold type?

src/stir_ldt/prover.rs
```diff
@@ -288,9 +209,128 @@ where
...
+        let merkle_proof = ctx
+            .merkle
+            .generate_multi_proof(r_shift_indexes.clone())
+            .unwrap();
+
+        ctx.merkle_proofs
+            .push((merkle_proof, virtual_evals.clone()));
+
+        if ctx.round_1 {
+            match self.0.fold_optimization {
```
WizardOfMenlo, line 296: Why are we doing this here? We should just commit to the interpolated values (as in the WHIR impl)

WizardOfMenlo (changes requested):
I have left a few comments. I don't get why the folding strategy changes from the WHIR case, we should match it as much as possible. 
Have not checked the verifier, will do it next pass.

jdodinh:
I adjusted the mistakes that were there from the initial PR.

WizardOfMenlo (approved):

PR #152, opened 2025-03-31
sumcheck single: improve compress fn with bench

tcoratger:
@WizardOfMenlo I noticed by profiling a full WHIR run, that the compress function of the sumcheck takes a lot of time in the process

image

So I optimized it with some tricks and using some benchmarks, I've the following results:

```shell
     Running benches/compress.rs (target/release/deps/compress-6a5bbc53d07e5ee2)
Gnuplot not found, using plotters backend
compress/4              time:   [39.437 µs 39.798 µs 40.137 µs]
                        change: [-6.0949% -4.5054% -3.0623%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild
compress/64             time:   [52.307 µs 53.149 µs 53.905 µs]
                        change: [-8.4101% -6.4306% -3.9277%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  5 (5.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
compress/1024           time:   [124.09 µs 125.34 µs 126.63 µs]
                        change: [-7.6221% -5.7180% -3.8935%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
compress/16384          time:   [460.49 µs 468.78 µs 477.53 µs]
                        change: [-5.3329% -4.1556% -2.9310%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  7 (7.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
compress/1048576        time:   [8.1681 ms 8.2809 ms 8.4067 ms]
                        change: [-21.395% -17.997% -14.602%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

     Running benches/expand_from_coeff.rs (target/release/deps/expand_from_coeff-22ef1c5bc5c4ce96)

running 0 tests
```

I have another optimization idea for a followup PR concerning rounds where the polynomials are smaller than a certain threshold, even in the parallel config, we should fallback to a sequential loop because for small contexts, the setup of the parallel setting is even heavier than the sequential calculation. I think that the parallelization is only useful for the first rounds and probably less afterwards (we need to benchmark to check).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,9 @@ rand = "0.8"
 clap = { version = "4.4.17", features = ["derive"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-spongefish = { git = "https://github.com/arkworks-rs/spongefish", features = ["arkworks-algebra"] }
+spongefish = { git = "https://github.com/arkworks-rs/spongefish", features = [
+    "arkworks-algebra",
+] }
 spongefish-pow = { git = "https://github.com/arkworks-rs/spongefish" }
 rayon = { version = "1.10.0", optional = true }
 thiserror = "2.0"
@@ -68,6 +70,10 @@ parallel = [
 rayon = ["dep:rayon"]
 asm = ["ark-ff/asm"]
 
+# [[bench]]
+# name = "expand_from_coeff"
+# harness = false
+
 [[bench]]
-name = "expand_from_coeff"
+name = "compress"
 harness = false
--- a/benches/compress.rs
+++ b/benches/compress.rs
@@ -1,0 +1,47 @@
+use ark_ff::fields::Field;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use whir::{
+    crypto::fields::Field64 as F,
+    poly_utils::{coeffs::CoefficientList, multilinear::MultilinearPoint},
+    sumcheck::SumcheckSingle,
+    whir::statement::{Statement, Weights},
+};
+
+fn bench_compress(c: &mut Criterion) {
+    let mut group = c.benchmark_group("compress");
+
+    for &size in &[4u64, 64, 1024, 1 << 14, 1 << 20] {
+        // Skip invalid cases
+        if size < 2 || !size.is_power_of_two() {
+            continue;
+        }
+
+        let num_vars = size.trailing_zeros() as usize;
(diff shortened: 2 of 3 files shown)


WizardOfMenlo:
Can you test with larger benches? I tried this out now and on my PC for 2^26 the main version seems to consistently be 0.1s faster in the e2e compared to the new implementation (7.2s vs 7.3s). 
Could be my machine as well tbh.

tcoratger:
Thanks for double checking this, indeed, I just ran the benchmark on 2^26 and the difference is almost invisible so there is something not correct. Let's not merge for the moment, I need to think further about this :)

tcoratger:
@WizardOfMenlo If you have a couple of mins, can you check this again? To be safe, I just kept the original parallel thing but I just added a fallback to sequential logic for small size. Here are the modifications I obtained for the small sizes:

```
Gnuplot not found, using plotters backend
compress/4              time:   [18.248 µs 18.488 µs 18.703 µs]
                        change: [-53.397% -52.578% -51.706%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  7 (7.00%) low mild
compress/64             time:   [28.545 µs 28.736 µs 28.923 µs]
                        change: [-48.563% -47.984% -47.370%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
compress/1024           time:   [81.364 µs 82.356 µs 83.314 µs]
                        change: [-37.820% -36.944% -36.116%] (p = 0.00 < 0.05)
                        Performance has improved.
compress/2048           time:   [190.40 µs 191.70 µs 193.06 µs]
                        change: [-17.762% -16.783% -15.674%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It probably won't play a big role because these rounds are small, but it's worth doing this small optimization in my opinion.

WizardOfMenlo (approved):
Happy to merge, we can have some passes later to improve perf

PR #154, opened 2025-04-02
add bench for whir end to end

tcoratger:
@WizardOfMenlo I know that making the `make_whir_things` function is not the perfect/ideal solution but here it is a way to have a good end to end benchmark without any effort so I think for now this is quite a convenient approach.


WizardOfMenlo:
We should update this to divan, and use a more realistic instance (similar to the one in whir_p3)

PR #157, opened 2025-04-02
Adding support for WHIR batching

yswami-tfh:
Changes:

1. New set of API: DomainSeparator::batch_commit_statement Committer::batch_commit Prover::batch_prove Verifier::verify_batched
2. New cargo feature flag "batching" to enable disable this API
3. New traits to make the verifier generic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ criterion = "0.5"
 debug = true
 
 [features]
-default = ["parallel"]
+default = ["parallel", "batching" ]
 parallel = [
     "dep:rayon",
     "ark-poly/parallel",
@@ -67,6 +67,7 @@ parallel = [
 ]
 rayon = ["dep:rayon"]
 asm = ["ark-ff/asm"]
+batching = []
 
 [[bench]]
 name = "expand_from_coeff"
--- a/src/parameters.rs
+++ b/src/parameters.rs
@@ -376,6 +376,8 @@ where
     ///
     /// These define the hashing function used when combining two child nodes into a parent node.
     pub two_to_one_params: TwoToOneParam<MerkleConfig>,
+
+    pub enable_batching: bool,
 }
 
 impl<MerkleConfig, PowStrategy> Display for WhirParameters<MerkleConfig, PowStrategy>
--- a/src/whir/batching.rs
+++ b/src/whir/batching.rs
@@ -1,0 +1,796 @@
+#![cfg(feature = "batching")]
+
+///
+/// High level idea of batching multiple polynomials is as follows:
+///
+/// - Prover commits to multiple Merkle roots of evaluations of each polynomial
+///
+/// - Verifier samples a batching_randomness field element. For soundness
+///   reasons, Fiat-Shamir batching randomness must be sampled after the Merkle
+///   roots have been committed.
+///
+/// - Prover computes the weighted sum of each individual polynomial based on
+///   batching_randomness to compute a single polynomial to proceed during
+///   Sumcheck rounds and any other future round.
+///
+/// - In the first round of the proof, Prover includes the STIR queries from
+///   each individual oracle. (The STIR query indexes should be the same for all
+///   oracles.)
+///
+/// - During proof verification, for the first round, the Verifier validates
+///   each individual STIR query in the proof against individual Merkle root.
+///   Once these Merkle paths are validated, the Verifier re-derives
+///   batching_randomness and combines the STIR responses using powers of
+///   batching_randomness.
+///
+/// - After the first round, rest of the protocol proceeds as usual.
+///
+use ark_crypto_primitives::merkle_tree::{
+    Config as MTConfig, LeafParam, MerkleTree, MultiPath, TwoToOneParam,
+};
+use ark_ff::{FftField, Field};
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+#[cfg(any(feature = "parallel", feature = "rayon"))]
+use rayon::prelude::*;
+use spongefish::{
+    codecs::arkworks_algebra::{
+        ByteDomainSeparator, ByteReader, ByteWriter, DeserializeField, FieldDomainSeparator,
+        FieldToUnit, UnitToField,
+    },
+    ProofError, ProofResult, UnitToBytes,
+};
+use spongefish_pow::PoWChallenge;
+
+use crate::{
+    fs_utils::{OODDomainSeparator, WhirPoWDomainSeparator},
+    poly_utils::coeffs::CoefficientList,
+    sumcheck::SumcheckSingleDomainSeparator,
+    whir::{
+        committer::{Committer, Witness},
+        domainsep::{DigestDomainSeparator, WhirDomainSeparator},
+        fs_utils::{DigestReader, DigestWriter},
+        parameters::WhirConfig,
+        prover::Prover,
+        statement::{Statement, StatementVerifier},
+        verifier::Verifier,
+        RoundZeroProofValidator, WhirCommitmentData, WhirProof, WhirProofRoundData,
+    },
+};
+
+fn fma_stir_queries<F: Field>(
+    scale_factor: F,
+    stir_queries: &[Vec<F>],
+    folded_queries: &mut [Vec<F>],
+) {
+    for (folded_vec, stir_vec) in folded_queries.iter_mut().zip(stir_queries.iter()) {
+        for (i, value) in stir_vec.iter().enumerate() {
+            folded_vec[i] += scale_factor * value
+        }
+    }
+}
+
+#[cfg(debug_assertions)]
+fn validate_stir_queries<F, MerkleConfig>(
+    batching_randomness: &F,
+    expected_stir_value: &[Vec<F>],
+    stir_list: &[(MultiPath<MerkleConfig>, Vec<Vec<F>>)],
+) -> bool
+where
+    F: Field,
+    MerkleConfig: MTConfig<Leaf = [F]>,
+{
+    if stir_list.is_empty() {
+        return true;
+    }
+
+    let (basic_indexes, mut computed_stir) = stir_list[0].clone();
+    let mut multiplier = *batching_randomness;
+
+    for (multi_path, stir_queries) in &stir_list[1..] {
+        if !multi_path
+            .leaf_indexes
+            .iter()
+            .zip(basic_indexes.leaf_indexes.iter())
+            .all(|(lhs, rhs)| lhs == rhs)
+        {
+            return false;
+        }
+
+        fma_stir_queries(multiplier, stir_queries.as_slice(), &mut computed_stir);
+        multiplier *= batching_randomness;
+    }
+
+    let result = expected_stir_value == computed_stir;
+    result
+}
+
+#[cfg(not(debug_assertions))]
+fn validate_stir_queries<F, MerkleConfig>(
+    _batching_randomness: &F,
+    _expected_stir_value: &[Vec<F>],
+    _stir_list: &[(MultiPath<MerkleConfig>, Vec<Vec<F>>)],
+) -> bool
+where
+    F: Field,
+    MerkleConfig: MTConfig<Leaf = [F]>,
+{
+    true
+}
+
+#[derive(Clone, CanonicalSerialize, CanonicalDeserialize)]
+pub struct BatchedWhirProof<MerkleConfig, F>
+where
+    MerkleConfig: MTConfig<Leaf = [F]>,
+    F: Sized + Clone + CanonicalSerialize + CanonicalDeserialize,
+{
+    ///
+    /// List of Merkle paths and the PCP slot value for each individual Merkle
+    /// tree in the batch. The order of MultiPath must match the order of roots
+    /// in the commitment.
+    ///
+    pub first_round_paths: Vec<(MultiPath<MerkleConfig>, Vec<Vec<F>>)>,
+
+    /// List of Merkle Paths after first batched round
+    pub merkle_paths: Vec<(MultiPath<MerkleConfig>, Vec<Vec<F>>)>,
+
+    /// Statement value at a random point.
+    pub statement_values_at_random_point: Vec<F>,
+}
+
+///
+/// This is the list of [Witness] structs as if each polynomial will be folded
+/// independently. The witnesses are individually committed to the FS-transcript.
+///
+pub struct BatchedWitness<F, MerkleConfig: MTConfig> {
+    pub(crate) witness_list: Vec<Witness<F, MerkleConfig>>,
+    pub(crate) batching_randomness: F,
+}
+
+impl<F, MerkleConfig> BatchedWitness<F, MerkleConfig>
+where
+    MerkleConfig: MTConfig,
+    F: Field,
+{
+    ///
+    /// Combines all polynomials and leaf nodes weighted by powers of
+    /// batching_randomness and returns the pair (batched_polynomial,
+    /// batched_oracle). This function is expensive O(2^num_variables).
+    ///
+    pub fn batch_witnesses(&self) -> (CoefficientList<F>, Vec<F>) {
+        assert!(
+            self.witness_list.len() > 1,
+            "Batched witness list should have at least 2 elements"
+        );
+
+        assert!(self.witness_list.iter().all(|wit| wit
+            .polynomial
+            .coeffs()
+            .len()
+            .is_power_of_two()
+            && wit.merkle_leaves.len().is_power_of_two()
+            && wit.polynomial.num_variables() == self.witness_list[0].polynomial.num_variables()
+            && wit.merkle_leaves.len() == self.witness_list[0].merkle_leaves.len()),);
+
+        let poly_dim = self.witness_list[0].polynomial.coeffs().len();
+        let oracle_len = self.witness_list[0].merkle_leaves.len();
+        assert!(oracle_len % poly_dim == 0);
+        let rate_factor = oracle_len / poly_dim;
+
+        let mut batched_poly: Vec<F> = self.witness_list[0].polynomial.coeffs().to_vec();
+        let mut batched_oracle: Vec<F> = self.witness_list[0].merkle_leaves.clone();
+
+        let mut multiplier = self.batching_randomness;
+
+        // This could potentially be parallelized.
+        for wit in &self.witness_list[1..] {
+            for (i, coefficient) in wit.polynomial.coeffs().iter().enumerate() {
+                batched_poly[i] += multiplier * coefficient;
+
+                for j in 0..rate_factor {
+                    let ndx = i + j * poly_dim;
+                    batched_oracle[ndx] += multiplier * wit.merkle_leaves[ndx];
+                }
+            }
+
+            multiplier *= self.batching_randomness;
+        }
+        (CoefficientList::new(batched_poly), batched_oracle)
+    }
+
+    pub fn roots(&self) -> Vec<MerkleConfig::InnerDigest> {
+        self.witness_list
+            .iter()
+            .map(|w| w.merkle_tree.root())
+            .collect()
+    }
+}
+
+pub trait BatchedDomainSeparator<F, MerkleConfig>: WhirDomainSeparator<F, MerkleConfig>
+where
+    F: FftField,
+    MerkleConfig: MTConfig,
+{
+    #[must_use]
+    fn batch_commit_statement<PowStrategy>(
+        self,
+        batch_size: u32,
+        params: &WhirConfig<F, MerkleConfig, PowStrategy>,
+    ) -> Self;
+}
+
+impl<F, MerkleConfig, DomainSeparator> BatchedDomainSeparator<F, MerkleConfig> for DomainSeparator
+where
+    F: FftField,
+    MerkleConfig: MTConfig,
+    DomainSeparator: ByteDomainSeparator
+        + FieldDomainSeparator<F>
+        + SumcheckSingleDomainSeparator<F>
+        + WhirPoWDomainSeparator
+        + OODDomainSeparator<F>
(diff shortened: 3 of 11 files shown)


Cargo.toml
```diff
@@ -60,7 +60,7 @@ criterion = "0.5"
 debug = true
 
 [features]
-default = ["parallel"]
+default = ["parallel", "batching" ]
```
recmo, line 63: I'm not sure if we need a feature flag for this. Generally features are used to gate dependencies and target specific functionality, not actual features :)

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+};
+
+fn fma_stir_queries<F: Field>(
+    scale_factor: F,
+    stir_queries: &[Vec<F>],
+    folded_queries: &mut [Vec<F>],
+) {
+    for (folded_vec, stir_vec) in folded_queries.iter_mut().zip(stir_queries.iter()) {
+        for (i, value) in stir_vec.iter().enumerate() {
+            folded_vec[i] += scale_factor * value
```
recmo, line 67: In Rust the convention is to avoid indices as much as possible. (This guarantees elimination of bounds checking and generally simplifies code)

```rust
for (folded, value) in folded_vec.iter_mut().zip(stir_vec.iter()) {
    *folded += scale_factor * value
```

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+
+        fma_stir_queries(multiplier, stir_queries.as_slice(), &mut computed_stir);
+        multiplier *= batching_randomness;
+    }
+
+    let result = expected_stir_value == computed_stir;
+    result
+}
+
+#[cfg(not(debug_assertions))]
```
recmo, line 107: I would apply the `#[cfg(debug_assertions)]` on the call site. Or use `debug_assert!` which has this built in.

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+        );
+
+        assert!(self.witness_list.iter().all(|wit| wit
+            .polynomial
+            .coeffs()
+            .len()
+            .is_power_of_two()
+            && wit.merkle_leaves.len().is_power_of_two()
+            && wit.polynomial.num_variables() == self.witness_list[0].polynomial.num_variables()
+            && wit.merkle_leaves.len() == self.witness_list[0].merkle_leaves.len()),);
```
recmo, line 172: It's best to keep `assert` expressions as simple as possible, and definitely break them down to individual clauses instead of using `and`. I learned this the hard way, getting back a crash report from a client that said "assertion failed: clause_1 && clause_2 && clause_3", and had to ship an update first to figure out which clause failed to diagonse the problem)

```rust
for witness in &self.witness_list {
  assert!(witness.coeffs().len().is_power_of_two());
  assert!(witness.merkle_leaves.len().is_power_of_two());
  // ...
}
```

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+
+        let poly_dim = self.witness_list[0].polynomial.coeffs().len();
+        let oracle_len = self.witness_list[0].merkle_leaves.len();
+        assert!(oracle_len % poly_dim == 0);
```
recmo, line 176: For this there is `assert_eq` which includes lhs and rhs values in crash output.

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+
+        // This could potentially be parallelized.
+        for wit in &self.witness_list[1..] {
+            for (i, coefficient) in wit.polynomial.coeffs().iter().enumerate() {
+                batched_poly[i] += multiplier * coefficient;
+
+                for j in 0..rate_factor {
+                    let ndx = i + j * poly_dim;
+                    batched_oracle[ndx] += multiplier * wit.merkle_leaves[ndx];
+                }
```
recmo, line 192: Is it possible to flip order of iteration so the inner loop is consecutive values? You can then uses `chunks_exact` on `batched_oracle` to avoid indices.

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+            "Batched evaluation requires more than one polynomial"
+        );
+
+        assert!(
+            polynomial_list.iter().all(|poly| {
+                poly.num_variables() == polynomial_list[0].num_variables()
+                    && poly.num_coeffs() == polynomial_list[0].num_coeffs()
+            }),
+            "Each polynomial must have same number of variables and same number of coefficients"
+        );
```
recmo, line 295: See above comment on conjuctive assetions.

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+    where
+        ProverFSState: FieldToUnitSerialize<F>
+            + BytesToUnitSerialize
+            + DigestToUnitSerialize<MerkleConfig>
+            + UnitToField<F>,
+    {
+        assert!(
+            polynomial_list.len() > 1,
+            "Batched evaluation requires more than one polynomial"
+        );
```
recmo, line 287: Can we make it such that for `len() == 1` it reproduces the normal unbatched behaviour?

The pattern I would expect here is 

```rust
commit(&self, prover_state: _, polynomial: &CoefficientList<_>) {
   self.commit(prover_state, &[polynomial])
}

commit_batch(&self, prover_state: _, polynomial_list: &[CoefficientList<F::BasePrimeField>]) {
  // ..
}
```

or 


```rust
commit(&self, prover_state: _, polynomial: &CoefficientList<_>) {
   // ..
}

commit_batch(&self, prover_state: _, polynomial_list: &[CoefficientList<F::BasePrimeField>]) {
  if let [polynomial] = polynomial_list {
     return self.commit(prover_state, polynomial);
  }

  // ..
}
```

which of the two is preferred depends on what is simpler to implement. If a batch of `len() == 1` naturally reproduces the unbatched behaviour (or with a few conditionals) then the first form is preferred. This would be best IMO as it results in the simples implementation. But if the two cases share little code, the 2nd form may be preferred.

recmo: This also implies that the batching versions of structs and non-batching versions are merged, which is something we should strongly consider as it simplifies the codebase.

src/whir/batching.rs
```diff
@@ -0,0 +1,819 @@
...
+        }
+
+        assert!(
+            validate_stir_queries(
+                &commitment.batching_randomness,
+                &batched_stir,
+                &first_round_paths,
+            ),
+            "The computed value of STIR queries must match the value in the leaf nodes"
+        );
```
recmo, line 361: Use a `debug_assert!` here and remove the conditional compilation on `validate_stir_queries`.

recmo (commented):
I'd like to consider a simpler approach where we add batching versions of functions to the existing structs. I suspect this will result in a simpler implementation as we can re-use more code, but I may also be missing something. @WizardOfMenlo WDYT?

WizardOfMenlo (commented):
I agree with @recmo, we should simplify things a bit. 
My proposal is this:

- Change the params to (instead of having a enable_batching flag) have a `batch_size` parameter
- This also allows to update the DomainSeparator properly
- The CommitmentWriter has a `commit_batch`  and a `commit` (which should just fall back on the commit_batch implementation)
- The CommitmentReader returns now `ParsedCommitment`. I originally thought that this would be an enum, but now I kind of thing it can be the same struct, only with a Vec of roots and a VecVec of OOD points and answers (open Q: can we handle the case where we batch many into a single merkle tree be handled smoothly here?)
- The ParsedCommitment should now have methods to generate the Merkle proofs smoothly (for the prover) and verify them (for the verifier).
What do you think?

yswami-tfh:
Generally agree with the comments. Let me unify the structs and remove unnecessary traits.

src/whir/batching.rs
yswami-tfh: Removed indexes. This code has now [moved to](fma_stir_queries) `whir::utils::fma_stir_queries`

src/whir/batching.rs
yswami-tfh: I'm not too familiar with caching related issues, so have left it as it is for the moment.

src/whir/batching.rs
yswami-tfh: Batching and non-batching structs all unified all across the crate. This change however is more intrusive than before (but this is unavoidable).

The only new function is `CommitmentWriter::batched_commit` Every other function such as `prove`, `verify` `parse` etc. work with both batched and non-batched operations.

yswami-tfh:
Have taken care of almost all the comments. Git-commit [8c6c996](https://github.com/WizardOfMenlo/whir/pull/157/commits/8c6c9965dc48436cb1bd36e34b8b6b9950764903) has all the changes in one place. The changes are more invasive so removed the cargo feature flag as well.

src/parameters.rs
```diff
@@ -433,6 +436,19 @@ where
     }
 }
 
+impl<MerkleConfig, PowStrategy> WhirParameters<MerkleConfig, PowStrategy>
+where
+    MerkleConfig: Config,
+{
+    pub(crate) fn rbr_soundness(&self, round: usize) -> SoundnessType {
+        if self.batch_size > 1 && round == 0 {
+            SoundnessType::UniqueDecoding
```
WizardOfMenlo, line 445: Why? We can be in list decoding even in the batching case?

src/whir/committer/mod.rs
```diff
@@ -8,6 +8,19 @@ pub mod writer;
 pub use reader::CommitmentReader;
 pub use writer::CommitmentWriter;
 
+///
+/// Represents the individual leaves  and merkle tree associated
+/// with the polynomial.
+///
+#[derive(Clone)]
+pub(crate) struct BatchingData<F, MerkleConfig>
```
WizardOfMenlo, line 16: I don't really get the point of this struct.
We should just have Witness that contains a Vec of merkle_trees and a Vec of merkle leaves no?

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
...
-    /// The committed polynomial in coefficient form.
+    /// The committed polynomial in coefficient form. In case of batching, its
+    /// the batched polynomial, i.e., the weighted sum of polynomials in
+    /// batching data.
     pub(crate) polynomial: CoefficientList<F>,
-    /// The Merkle tree constructed from the polynomial evaluations.
+
+    /// The Merkle tree constructed from the polynomial evaluations. In case of
+    /// batching, it's the merkle tree of the batched polynomial.
     pub(crate) merkle_tree: MerkleTree<MerkleConfig>,
```
WizardOfMenlo, line 40: This becomes a Vec<MT>

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
 where
     MerkleConfig: Config,
 {
-    /// The committed polynomial in coefficient form.
+    /// The committed polynomial in coefficient form. In case of batching, its
+    /// the batched polynomial, i.e., the weighted sum of polynomials in
+    /// batching data.
     pub(crate) polynomial: CoefficientList<F>,
```
WizardOfMenlo, line 36: This becomes a Vec<polynomial>

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
...
+    /// evaluations. In case of batching, its the merkle leaves of the batched
+    /// tree. These leaves are computed as the weighted sum leaf values in the
+    /// batching_data.
     pub(crate) merkle_leaves: Vec<F>,
+
     /// Out-of-domain challenge points used for polynomial verification.
     pub(crate) ood_points: Vec<F>,
+
     /// The corresponding polynomial evaluations at the OOD challenge points.
     pub(crate) ood_answers: Vec<F>,
```
WizardOfMenlo, line 52: This becomes a Vec<Vec<F>> containing the OOD answers to each polynomial

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
...
+
+    /// The individual commitments of each batching polynomial
+    pub(crate) batching_data: Vec<BatchingData<F, MerkleConfig>>,
```
WizardOfMenlo, line 55: This goes

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
...
+
+    /// The batching randomness. If there's no batching, this value is zero.
+    pub(crate) batching_randomness: F,
```
WizardOfMenlo, line 58: This can become a Vec<F>, in the batching_1 this will be set to [F::ONE], otherwise it's [F::ONE, r, r^2, ...] ecc.

src/whir/committer/mod.rs
```diff
@@ -17,14 +30,42 @@ pub struct Witness<F, MerkleConfig>
...
+}
+
+impl<F, MerkleConfig> From<Witness<F, MerkleConfig>> for BatchingData<F, MerkleConfig>
```
WizardOfMenlo, line 61: Do we need this?

src/whir/committer/reader.rs
```diff
@@ -2,16 +2,28 @@ use ark_crypto_primitives::merkle_tree::Config;
...
+/// the number of virtual oracles and the value of batching_randomness will
+/// neither be F::ZERO nor F::ONE (w.h.p.).
+///
+
 #[derive(Clone)]
 pub struct ParsedCommitment<F, D> {
-    pub root: D,
+    pub root: Vec<D>,
     pub ood_points: Vec<F>,
     pub ood_answers: Vec<F>,
```
WizardOfMenlo, line 25: This becomes `Vec<Vec<F>>`

src/whir/committer/reader.rs
```diff
@@ -2,16 +2,28 @@ use ark_crypto_primitives::merkle_tree::Config;
...
+    pub batching_randomness: F,
```
WizardOfMenlo, line 26: Same as in `Witness`

src/whir/committer/reader.rs
```diff
@@ -35,6 +47,28 @@ where
...
         verifier_state: &mut VerifierState,
     ) -> ProofResult<ParsedCommitment<F, MerkleConfig::InnerDigest>>
+    where
+        VerifierState: UnitToBytes
+            + FieldToUnitDeserialize<F>
+            + UnitToField<F>
+            + DigestToUnitDeserialize<MerkleConfig>
+            + BytesToUnitDeserialize,
+    {
+        let mut batch_size_le_bytes: [u8; 4] = [0, 0, 0, 0];
```
WizardOfMenlo, line 57: This should not be read from the transcript, it's part of the protocol parameters so we can read it from there.

src/whir/committer/reader.rs
```diff
@@ -35,6 +47,28 @@ where
...
+        verifier_state.fill_next_bytes(&mut batch_size_le_bytes)?;
+        let batch_size = u32::from_le_bytes(batch_size_le_bytes) as usize;
+
+        if batch_size == 1 {
```
WizardOfMenlo, line 61: Can we not have a single function instead of two, seems like a lot of duplication

src/whir/committer/reader.rs
```diff
@@ -49,9 +83,48 @@ where
...
+            + BytesToUnitDeserialize,
+    {
+        let mut roots = Vec::<MerkleConfig::InnerDigest>::with_capacity(batch_size);
+
+        for _ in 0..batch_size {
+            // Order is important here
+            let root = verifier_state.read_digest()?;
+            roots.push(root);
+        }
+
```
WizardOfMenlo, line 112: MIssing the OOD samples??

src/whir/batching.rs
WizardOfMenlo, line 1: Do we need this file? I would be happy to keep the tests, but ideally the first round branching happens in the verifier (if at all)

src/whir/committer/writer.rs
```diff
@@ -50,7 +52,7 @@ where
     /// - Computes out-of-domain (OOD) challenge points and their evaluations.
     /// - Returns a `Witness` containing the commitment data.
     #[cfg_attr(feature = "tracing", instrument(skip_all, fields(size = polynomial.num_coeffs())))]
-    pub fn commit<ProverState>(
+    fn commit_one<ProverState>(
```
WizardOfMenlo, line 55: `commit_single`

src/whir/committer/writer.rs
```diff
@@ -127,8 +128,168 @@ where
             merkle_leaves: folded_evals,
             ood_points,
             ood_answers,
+            batching_data: Vec::new(),
+            batching_randomness: F::ZERO,
         })
     }
+
+    pub fn batch_commit<ProverFSState>(
```
WizardOfMenlo, line 136: `commit_batch`

src/whir/committer/writer.rs
```diff
@@ -127,8 +128,168 @@ where
...
+        ProverFSState: FieldToUnitSerialize<F>
+            + BytesToUnitSerialize
+            + DigestToUnitSerialize<MerkleConfig>
+            + UnitToField<F>,
+    {
+        assert_eq!(polynomial_list.len(), self.0.batch_size);
+
+        assert!(polynomial_list.len() > 0,);
+
+        // 1. Add the batch size as u32 little endian to fs-transcript
```
WizardOfMenlo, line 151: No, we should not.

src/whir/committer/writer.rs
```diff
@@ -127,8 +128,168 @@ where
...
+        let batch_size = u32::try_from(polynomial_list.len()).unwrap().to_le_bytes();
+        prover_state.add_bytes(&batch_size)?;
+
+        if polynomial_list.len() == 1 {
```
WizardOfMenlo, line 155: We should aim to avoid this branching whenever possible.

src/whir/domainsep.rs
```diff
@@ -32,17 +32,44 @@ where
     DomainSeparator:
         ByteDomainSeparator + FieldDomainSeparator<F> + DigestDomainSeparator<MerkleConfig>,
 {
+    //
+    // FS Batch Commitment:
+    // P -> V
+    //  32-bit (4-bytes-le): Number of entries in the batch
```
WizardOfMenlo, line 38: Remove, it's in parameters (which will be absorbed into the FS, we are not doing it yet here)

WizardOfMenlo (changes requested):
I have done a preliminary pass. There is a few things I am not happy with, so would not merge right now.
- There are two cases in which we might want to do a batch WHIR proof, and I think we are supporting neither of them properly:
    1. The WHIR ZK case. Here the flow of the protocol is roughly: (i) prover commits to (masked) poly, (ii) things happen (iii) prover samples and commit to a sumcheck masking polynomial (iv) verifier samples batching randomness and then we run the batched WHIR. Currently, I don't see this being supported in the API.
    2. The WHIR succinct case where you have a bunch of polynomials committed to in the same round of the protocol. This seems to be what `commit_batch` supports, but in this case we should commit with a single Merkle tree instead of what we are doing now.
- The batch_size should not appended to the transcript. It's part of the parameters and that is enough (the parameters should also be added to the DomainSep, but that something I haven't done yet myself).
- Overall, there is a lot of branching that looks very fishy and can probably be simplified.

Thinking of a nice API to do both case 1 and 2 would go a long way. Let me know if you agree @yswami-tfh @recmo

src/whir/committer/mod.rs
yswami-tfh: There's a fundamental difference between batching and non-batching case. Without batching there's  a single polynomial and a single Merkle tree. With batching, say 2, there are two polynomials, 2 Merkle trees, and a third combined polynomial and Merkle tree over which the entire protocol runs. The `BatchinData` struct keeps the individual Merkle trees, while the combined polynomial and combined Merkle tree is part of the `Witness`. 

If we put all of this in a single vec, that will be more cumbersome in my opinion.

src/whir/committer/writer.rs
yswami-tfh: If we don't put the batch side then parsing the FS-transcript not will work without support from the user.

src/parameters.rs
yswami-tfh: This is only for the first round (round == 0). The tradeoff here is that we either add OOD in the commitment or we do UniqueDecoding in the first round. The OOD will have to have the same indexes for each top-level oracle, which sort of defeats the purpose of OOD in my opinion, so my preference was for doing unique decoding over OOD (but don't have any strong opinions about it :-) )

yswami-tfh: Added OOD support for batching and removed this check.

src/whir/committer/mod.rs
yswami-tfh: Its done.

src/whir/committer/writer.rs
yswami-tfh: Done

src/whir/committer/writer.rs
yswami-tfh: Done.

src/whir/committer/writer.rs
yswami-tfh: Done.

src/whir/domainsep.rs
yswami-tfh: Done.

src/whir/committer/mod.rs
yswami-tfh: I think it's better to have custom data structures that group related things together than use multiple independent `Vec`s.

src/whir/committer/writer.rs
yswami-tfh: Done.

src/whir/batching.rs
yswami-tfh: Moved to `Prover.rs`

PR #175, opened 2025-05-15
Add batching support.

recmo:


recmo:
I actually can push to #157 , so will continue there.

PR #176, opened 2025-05-17
Use hints for MerkleProof

recmo:
Depends on a corresponding feature in Spongefish:

https://github.com/arkworks-rs/spongefish/pull/57


src/crypto/merkle_tree/parameters.rs
```diff
@@ -84,6 +84,15 @@ where
...
 }
 
+impl HintSerialize for ProverState {
+    fn hint<T: CanonicalSerialize>(&mut self, hint: &T) -> ProofResult<()> {
+        let mut bytes = Vec::new();
+        hint.serialize_compressed(&mut bytes)?;
+        self.hint_bytes(&bytes)?;
+        Ok(())
+    }
+}
```
recmo, line 94: It would make sense to move these to `spongefish`.

src/whir/mod.rs
```diff
@@ -16,39 +15,30 @@ pub mod verifier;
...
-pub struct WhirProof<MerkleConfig, F>
+pub struct WhirProof<F>
 where
-    MerkleConfig: Config<Leaf = [F]>,
     F: Sized + Clone + CanonicalSerialize + CanonicalDeserialize,
 {
-    #[serde(with = "crate::ark_serde")]
-    pub merkle_paths: Vec<(MultiPath<MerkleConfig>, Vec<Vec<F>>)>,
     #[serde(with = "crate::ark_serde")]
     pub statement_values_at_random_point: Vec<F>,
```
recmo, line 23: I don't think this should be here either, but I'll leave it for a follow up PR to remove. I think it's the verifiers responsibility to compute this.

recmo: @veljkovranic Why is `statement_values_at_random_point` part of the proof? Shouldn't verifier compute this?

recmo: I moved these to verifier work, but for a practical verifier I'll have to re-add a method to defer or precompute the evaluation and not force a full evaluation of the MLE in the verifier.

Cargo.toml
```diff
@@ -44,10 +44,14 @@ rand = "0.8"
 clap = { version = "4.4.17", features = ["derive"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-spongefish = { git = "https://github.com/arkworks-rs/spongefish", features = [
+# spongefish = { git = "https://github.com/arkworks-rs/spongefish", features = [
+#     "arkworks-algebra",
+# ] }
+# spongefish-pow = { git = "https://github.com/arkworks-rs/spongefish" }
+spongefish = { git = "https://github.com/recmo/nimue", branch = "remco/hint", features = [
```
WizardOfMenlo, line 51: Revert to arkworks? Unless we have other changes to incorporate

src/whir/domainsep.rs
```diff
@@ -88,7 +90,10 @@ where
 
         self.add_scalars(1 << params.final_sumcheck_rounds, "final_coeffs")
             .challenge_bytes(domain_size_bytes * params.final_queries, "final_queries")
+            .hint("stir_answers")
+            .hint("merkle_proof")
             .pow(params.final_pow_bits)
             .add_sumcheck(params.final_sumcheck_rounds, params.final_folding_pow_bits)
+            .hint("deferred")
```
WizardOfMenlo, line 97: weight_evaluation?

src/whir/parsed_proof.rs
```diff
@@ -60,8 +60,9 @@ pub(crate) struct ParsedProof<F> {
     pub(crate) final_sumcheck_randomness: MultilinearPoint<F>,
     /// Coefficients of the final small polynomial.
     pub(crate) final_coefficients: CoefficientList<F>,
-    /// Evaluation values of the statement being proven at a random point.
-    pub(crate) statement_values_at_random_point: Vec<F>,
+
+    /// Deferred constraint evaluations.
+    pub(crate) deferred: Vec<F>,
```
WizardOfMenlo, line 65: weight_evaluation

src/whir/statement.rs
```diff
@@ -181,6 +189,10 @@ pub struct Constraint<F> {
     pub weights: Weights<F>,
     #[serde(with = "crate::ark_serde")]
     pub sum: F,
+
+    /// For a deferred statement the verifier ignores the weights and
+    /// instead outputs a relation on them that should be verified.
+    pub deferred: bool,
```
WizardOfMenlo, line 195: As before

src/whir/utils.rs
```diff
@@ -91,6 +92,14 @@ pub trait DigestToUnitDeserialize<MerkleConfig: Config> {
     fn read_digest(&mut self) -> ProofResult<MerkleConfig::InnerDigest>;
 }
 
+pub trait HintSerialize {
```
WizardOfMenlo, line 95: I think it would maybe be cleaner to have an hint trait for each type we want to serialize. This way at call site we can understand that we are doing the right hint addition

WizardOfMenlo (approved):
Happy to merge, maybe let's change `deferred` to something like `weights_evaluations` to be more descriptive.
I don't really like the blanket canonical serialize trait (for HintSerialize and HintDeserialize), I somehow would like to see `hint_merkle_proof`, `hint_field_elements` ecc.

recmo:
I added explicit types in the prover for hints:

```rust
        prover_state.hint::<Vec<Vec<F>>>(&answers)?;
        prover_state.hint::<MultiPath<MerkleConfig>>(&merkle_proof)?;
```

This is a little bit more verbose than introducing `hint_merkle_proof` and `hint_field_elements`, but it does not require introducing a new trait + impl for every type. Plus it contains the precise type instead of a description.


Yet I'm also not satisfied with this. I'd like there to be more checks that prover and verifier are doing the exact mirror image:

- It should be an error if the prover and verifier type are not the same.

- Currently there are only labels in the domainseparator, but not in prover/verifier so there is no verification that labels match.

- There is no modularity in the sense of a 'begin_whir_proof' and 'end_whir_proof' marker. This makes the errors confusing if the whir proof forgets a term, as the error will happen at the next interaction outside of the whirproof.

PR #179, opened 2025-05-20
Modular proofs

recmo:


recmo:
Moving the transcript ideas to Spongefish.

PR #180, opened 2025-05-22
statement: enable parallelization for statement evaluate

tcoratger:
Similarly to what is done for the `weighted_sum` function

https://github.com/WizardOfMenlo/whir/blob/bd6b47818cfe197565d816ba1249060e22c5af32/src/whir/statement.rs#L149-L172

we can enable parallelization for looping over the weights.

### DIFF
--- a/src/whir/statement.rs
+++ b/src/whir/statement.rs
@@ -77,12 +77,25 @@ impl<F: Field> Weights<F> {
             Self::Evaluation { point } => poly.evaluate(point),
             Self::Linear { weight } => {
                 let poly: EvaluationsList<F> = poly.clone().into();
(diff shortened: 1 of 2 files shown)


WizardOfMenlo:
I am not sure we should merge this, as far as I can tell the only place this is called is within the verifier, which I was hoping to keep serial as much as possible.

tcoratger:
> I am not sure we should merge this, as far as I can tell the only place this is called is within the verifier, which I was hoping to keep serial as much as possible.

You're right, I hadn't noticed. I just updated the PR with a comment explaining why we don't use parallelization here so we don't make any mistakes in the future.

PR #182, opened 2025-05-30
Modular protocol

recmo:


recmo:
This is now fully superseded by #209, #210, #212 and #215.

PR #183, opened 2025-05-30
Interleaved RS Code for ProverHelps optimization. 

yswami-tfh:
Add interleaved RS encoding of polynomial
- Interleaved RS encoding using `rs_encode_coset_poly`.
- Moved old functions [expand_from_coeff](https://github.com/WizardOfMenlo/whir/blob/654326d096b6dc9ca21897e5e9ad20bfa4f9cd18/src/ntt/mod.rs#L24) and [transform_evaluations](https://github.com/WizardOfMenlo/whir/blob/654326d096b6dc9ca21897e5e9ad20bfa4f9cd18/src/poly_utils/fold.rs#L98) to tests
- Updated benchmark code
- Added better documentation

### DIFF
--- a/benches/expand_from_coeff.rs
+++ b/benches/expand_from_coeff.rs
@@ -4,10 +4,14 @@ use whir::{crypto::fields::Field64, ntt};
 #[global_allocator]
 static ALLOC: AllocProfiler = AllocProfiler::system();
 
-// Test cases with polynomial sizes defined as exponents of 2 and expansion factors
+//
+// Test cases with tuple entries:
+//      - polynomial sizes defined as exponents of 2,
+//      - RS code expansion factors, and
+//      - interleaved bloc size exponent of 2
+//
 const TEST_CASES: &[(u32, usize, usize)] = &[
     (16, 2, 2),
-    (18, 2, 2),
     (18, 2, 2),
     (20, 2, 3),
     (16, 4, 3),
@@ -17,7 +21,7 @@ const TEST_CASES: &[(u32, usize, usize)] = &[
 ];
 
 #[divan::bench(args = TEST_CASES)]
-fn rs_encode_coset_poly(bencher: Bencher, case: &(u32, usize, usize)) {
+fn interleaved_rs_encode(bencher: Bencher, case: &(u32, usize, usize)) {
     bencher
         .with_inputs(|| {
             let (exp, expansion, coset_sz) = *case;
@@ -27,7 +31,7 @@ fn rs_encode_coset_poly(bencher: Bencher, case: &(u32, usize, usize)) {
             (coeffs, expansion, coset_sz)
         })
         .bench_values(|(coeffs, expansion, coset_sz)| {
-            black_box(ntt::rs_encode_coset_poly(&coeffs, expansion, coset_sz))
+            black_box(ntt::interleaved_rs_encode(&coeffs, expansion, coset_sz))
         });
 }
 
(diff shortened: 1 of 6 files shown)


benches/expand_from_coeff.rs
```diff
@@ -5,27 +5,30 @@ use whir::{crypto::fields::Field64, ntt};
 static ALLOC: AllocProfiler = AllocProfiler::system();
 
 // Test cases with polynomial sizes defined as exponents of 2 and expansion factors
```
WizardOfMenlo, line 7: We should update this comment

src/ntt/mod.rs
```diff
@@ -19,56 +17,111 @@ pub use self::{
...
+///
+/// f(X) = h₀(X^k) + X·h₁(X^k) + ⋯ + X^{k-1}·h_{k-1}(X^k)
+///
+/// where each hₜ(X) is polynomial of maximum degree `N/k - 1`.
+///
+/// If ζ is the N-th root of unity, then [rs_encode_coset_poly] computes
+/// the Reed-Solomon code for each hₜ(x) using `ω = ζ^k` as the root of
+/// unity, and for each index i, places [h₀(ω^i), h₁(ω^i), ⋯
+/// ,h_{k-1}(ω^i)] consecutively in the output buffer.
+///
```
WizardOfMenlo, line 37: Alternatively, a better way of thinking about this (I think) is that this is encoding a message f = (f_1, ..., f_n) as the interleaved Reed solomon codeword 
i.e. you take (f_1, f_k, f_2k, ...), (f_2, ...) and you encode each one of them with a RS code of size n/k.
Maybe we should update this comment and the naming of the function appropriately (how about `interleaved_rs_encode`?)

Let's also maybe leave a TODO on whether it makes sense to store the polynomial already transposed (i.e. poly as (f_1, f_k, f_2k,..., f_2, f_2 +k, ...))
This might save us some of the transposes? Not sure if in other places this helps or hurts

src/poly_utils/fold.rs
```diff
@@ -71,96 +67,97 @@ pub fn compute_fold<F: Field>(
...
-            for v in answers.iter_mut() {
-                *v *= scale;
-                scale *= &*offset;
-            }
-        });
-}
-
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
```
WizardOfMenlo, line 71: chore: maybe let's move the transform evaluations to a test_utils file that we make pub(crate), I kind of like unit tests in file not to export anything

src/poly_utils/fold.rs
```diff
@@ -71,96 +67,97 @@ pub fn compute_fold<F: Field>(
...
     use crate::{
         crypto::fields::Field64,
-        ntt::transpose,
+        ntt::{intt_batch, transpose},
         poly_utils::{coeffs::CoefficientList, multilinear::MultilinearPoint},
     };
 
     type F = Field64;
 
+    /// Applies a folding transformation to evaluation vectors in-place.
```
WizardOfMenlo, line 85: Let's add some note that this is only used to test consistency with new commit functionality

WizardOfMenlo (approved):
Looks great! Left a few small comments to update docs but LGTM!

src/poly_utils/fold.rs
yswami-tfh: Will move both the two old function to`test_utils` file. (Those two functions have some interesting ideas, so didn't want to remove those functions entirely.)

WizardOfMenlo (approved):

PR #199, opened 2025-10-28
Reed Solomon Trait

xrvdg:
This PR adds a trait that allows for changing the Reed Solomon implementation. An use case for this is to allow for Reed Solomon/NTT implementations that are specialised for a particular field. 

The design here takes a vTable approach as in #182. Here the builtin interleaved RS encode is available via the tag `RSDefault`. 

In case a generic approach is preferred over vTable just the first commit can be merged instead. 

In a follow up PR support will be added for RS encodings whose result is in reverse-bit order or any other non-normal order.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -90,173 +90,63 @@ fn main() {
         args.pow_bits = Some(default_max_pow(args.num_variables, args.rate));
     }
 
+    runner(&args, field, merkle);
+}
+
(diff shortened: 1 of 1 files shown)


WizardOfMenlo:
Seems good to me! Small refactoring, can we add the RS configuration to the config instead of adding it as an extra parameter to commit and prove? Also, I think it would be nice for the verifier to also be aware of it, as different encodings will also change the verifier behaviour

PR #206, opened 2025-12-10
chore: separate the code in three different crates

dartdart26:
Separate to the `prover`, `common`, `verifier` and `cli`.

Rationale is that we'd like to be able to run the verifier without randomness, say on onchain targets (Solana SBF, WASM-based chains, etc.). Use the `custom` feature for getrandom in order to silence the linker when building for targets without support as the verifier doesn't need randomness anyway.

Add an `onchain` Cargo profile that can be used to compile such targets.

Also support cases in NTT where usize::BITS is not 64 for 2 adicity (e.g. WASM).

To test some of these targets, go to the verifier dir and:
```
CARGO_PROFILE_RELEASE_LTO=fat \
CARGO_PROFILE_RELEASE_DEBUG=false \
CARGO_PROFILE_RELEASE_STRIP=symbols \
CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1 cargo build-sbf
```

or

```
cargo build --target wasm32-unknown-unknown
```


WizardOfMenlo:
Let's close this for now, I've opened #253 to remember

PR #209, opened 2026-01-13
Update spongefish

recmo:
Updates Spongefish to a recent commit that includes the API refactor.

This also change Proof of Work to a dyn-trait based approach. This makes it easier for library users to add new algorithms and avoids having to weave a generic type like  `PowStrategy` all through the code.

Spongefish has deprecated the concept of hints which is critical for the performance of proofs involving Merkle Trees. To work around this a `transcript` module is introduced that wraps Spongefish and add the missing functionality.

It also introduces the idea of sub protocol modules, which have their own `Config` structs containing a `prove` and `verify` member taking a `&mut ProverState` and `&mut VerifierState` respectively. The `proof_of_work` and `sumcheck` module have been adapted to use this pattern.

In further PRs I'll introduce a similar `Config` + `ENGINES` pattern for Merkle trees and RS encoding (so it's easy to provide custom implementations) and further refactor `WhirConfig` into nested configs of sub protocols.

### DIFF
--- a/src/crypto/proof_of_work/none_engine.rs
+++ b/src/crypto/proof_of_work/none_engine.rs
@@ -6,7 +6,7 @@ use super::Engine;
 use crate::transcript::{Protocol, ProtocolId};
 
 pub const NONE: ProtocolId = ProtocolId::new(hex!(
-    "3a777695dad25e7b8a5e69f2ecf25e6a0f8927fd334377d297644b14a761dd3d"
+    "c2f551bf887c9a5db24d1da57246e5826ce7ea8bd20e072f5ec12f4616c04a50"
 ));
 
 #[derive(Clone, Copy, PartialEq, Eq, Default)]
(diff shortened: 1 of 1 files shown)


recmo:
This also change Proof of Work to a dyn-trait based approach. This makes it easier for library users to add new algorithms and avoids having to weave a generic type like  `PowStrategy` all through the code.

Spongefish has deprecated the concept of hints which is critical for the performance of proofs involving Merkle Trees. To work around this a `transcript` module is introduced that wraps Spongefish and add the missing functionality.

It also introduces the idea of sub protocol modules, which have their own `Config` structs containing a `prove` and `verify` member taking a `&mut ProverState` and `&mut VerifierState` respectively. The `proof_of_work` and `sumcheck` module have been adapted to use this pattern.

In further PRs I'll introduce a similar `Config` + `ENGINES` pattern for Merkle trees and RS encoding (so it's easy to provide custom implementations) and further refactor `WhirConfig` into nested configs of sub protocols.

PR #215, opened 2026-02-03
Use IRS commit for internal rounds.

recmo:
- [x] Move sumcheck configs to WhirConfig.
- [x] Use irs_commit for round commitments.
- [x] Remove redundant values from WhirConfig.
- [x] Change all `.zip` to `zip_strict`.
- [x] See if we can un-special case the initial and final round a bit more.
- [x] Simplify sumcheck.
- [x] Allow mixed weight evaluation so we don't have to lift polynomials.
- [x] Instead associating statements to polynomials and computing the cross matrix internally, have WHIR take in polynomials, constraints and a matrix of claimed values. This way, if many polynomials need to be evaluated in the same point (presumably a common scenario), then the constraint only appears once.
- [x] Clean up verifier.
- [x] Fixed pow per round in sumcheck.
- [x] Initial sumcheck iff initial constraints.
- [x] Cleanup WhirConfig.
- [x] Move `whir` into `protocols`.
- [x] Add back configurable NTT backend, but using TypeMap.
- [x] Keep peformance parity.

### Later
- Prover keeps RS-encoded, evaluation basis and basis versions of polynomial around. Reduce this redundancy.
- Trait for Weights, have whir take `&[&dyn Weight<F>]`.
- Use explicit sizes instead of `num_variables` where meaningful.
- Consider flipping sumcheck variable order to allow inplace updates.
- Create a trait for constraints.

### Someday
- Embedding support in sumcheck.
- Support arbitrary embeddings (not just basefield).
- Use a proper tensor abstraction, consider `ndarray`.
- Consider generating OOD points on opening, so they are homogenous.
- Add multi-commit openings to `irs_commit` proptest.
- Add proptest based testing for whir proofs.
- Make it work for non-power-of-two polynomials.
- Create an algevra for constraints.
- Add scopes (begin / end) to typed debug transcript.
- Add sponge consistency checks to debug transcript, to detect desync immediately.

### DIFF
--- a/src/algebra/ntt/mod.rs
+++ b/src/algebra/ntt/mod.rs
@@ -101,7 +101,7 @@ pub fn interleaved_rs_encode<F: 'static>(
 /// This function computes the RS-code for each interleaved message and
 /// outputs the interleaved alphabets in the same order as the input.
 ///
-#[cfg_attr(feature = "tracing", instrument(skip(coeffs), fields(size = coeffs.len())))]
+#[cfg_attr(feature = "tracing", instrument(level = "debug", skip(coeffs), fields(size = coeffs.len())))]
 fn ark_ntt<F: FftField>(coeffs: &[F], expansion: usize, interleaving_depth: usize) -> Vec<F> {
     assert!(expansion > 0);
     assert!(coeffs.len().is_multiple_of(interleaving_depth));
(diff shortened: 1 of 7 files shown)


src/algebra/sumcheck.rs
```diff
@@ -0,0 +1,57 @@
+use ark_ff::Field;
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
+/// Compute
```
Bisht13: Doc: The doc comment is truncated at `/// Compute`. Should describe the return value: (c0, c2) are the constant and quadratic coefficients of the sumcheck polynomial h(X) = c0 + c1·X + c2·X².

Bisht13 (approved):

PR #218, opened 2026-02-13
Reduce tracing noise from internal spans

Bisht13:
## Summary

- Demote internal implementation spans (`merkle_tree::layer`, `ntt::ark_ntt`, `coeffs::fold`, `weights::accumulate`, `fold::compute_fold`) to `debug` level so they don't show up in default INFO-level output.
- Switch noisy `instrument` attributes to `skip_all` on `sumcheck::prove`, `sumcheck::verify`, `proof_of_work::prove`, and `merkle_tree::open` to avoid dumping large field element arrays and full `Config` structs into span fields.

Tested against ProveKit's `complete_age_check` circuit — prove output dropped from 1261 to 274 lines with no change in behavior.


recmo (approved):
Thanks!

PR #219, opened 2026-02-14
Feat : added WHIR protocol 2.0 implementation#216

ocdbytes:


src/protocols/whir_zk/prover.rs
```diff
@@ -0,0 +1,496 @@
...
+    /// 1. Blinding with g to form P = ρ·f + g
+    /// 2. Running WHIR rounds on P with a virtual oracle L = ρ·f̂ + h
+    /// 3. Proving helper polynomial evaluations so verifier can reconstruct L
+    #[cfg_attr(feature = "tracing", instrument(skip_all, fields(num_polynomials = polynomials.len())))]
+    pub fn prove_zk<H, R>(
+        &self,
+        prover_state: &mut ProverState<H, R>,
+        polynomials: &[&CoefficientList<F::BasePrimeField>],
+        witness: &ZkWitness<F>,
+        helper_config: &Config<F>,
```
recmo: Can we introduce a dedicated `whir_zk::Config` that contains both `self` and `helper_config`?

The way the `protocol::*` modules are designed is that they are all configured through a  `module::Config` struct, (with the only exceptions being the two challenge helpers because they are so simple, but may even those should).

Another reason to do it is that we shouldn't extend `whir::Config` outside of that module.

src/protocols/whir_zk/utils.rs
```diff
@@ -0,0 +1,427 @@
...
+    /// Chosen such that 2^ℓ > conservative query upper bound
+    pub ell: usize,
+
+    /// μ: Number of variables in the witness polynomial
+    pub mu: usize,
+}
+
+impl ZkParams {
+    /// Compute ell and mu from WHIR parameters.
+    pub fn from_whir_params<F: FftField>(whir_params: &Config<F>) -> Self {
```
recmo: Same

src/protocols/whir_zk/utils.rs
```diff
@@ -0,0 +1,427 @@
...
+            msk,
+            g0_hat: g0,
+            m_poly,
+            g_hats,
+            params,
+        }
+    }
+
+    /// Extend msk to μ variables by padding with zeros
+    pub fn extend_msk(&self) -> CoefficientList<F> {
```
recmo: This function is unused?

src/protocols/whir_zk/config.rs
```diff
@@ -0,0 +1,152 @@
...
+};
+
+#[derive(Clone)]
+pub struct ZkParams {
+    /// Number of variables for helper polynomials (ℓ in the paper).
+    /// Chosen such that 2^ℓ > conservative query upper bound.
+    pub num_helper_variables: usize,
+
+    /// Number of variables in the witness polynomial (μ in the paper).
+    pub num_witness_variables: usize,
```
recmo: This is redundant. Use `config.main.initial_num_variables()` or something similar instead.

src/protocols/whir_zk/config.rs
```diff
@@ -0,0 +1,152 @@
...
+    algebra::fields::FieldWithSize,
+    parameters::{FoldingFactor, MultivariateParameters, ProtocolParameters, SoundnessType},
+    protocols::whir,
+};
+
+#[derive(Clone)]
+pub struct ZkParams {
+    /// Number of variables for helper polynomials (ℓ in the paper).
+    /// Chosen such that 2^ℓ > conservative query upper bound.
+    pub num_helper_variables: usize,
```
recmo: Redundant, use `config.helper.initial_num_variables()` instead.

src/protocols/whir_zk/config.rs
```diff
@@ -0,0 +1,152 @@
...
+/// Configuration for the ZK WHIR protocol.
+///
+/// Bundles the main WHIR config (for the witness polynomial), the helper
+/// WHIR config (for helper polynomials committed in a separate Merkle tree),
+/// and the derived ZK parameters.
+pub struct Config<F: FftField> {
+    /// Main WHIR config for the witness polynomial.
+    pub main: whir::Config<F>,
+    /// Helper WHIR config for helper polynomials (ℓ+1 variables, batch_size = N×(μ+1)).
+    pub helper: whir::Config<F>,
```
recmo: `helper` is not a helpful name. Please make sure to give all identifiers descriptive names.

src/protocols/whir_zk/verifier.rs
```diff
@@ -0,0 +1,628 @@
...
+    /// RLC coefficients combining multiple committed polynomials
+    polynomial_rlc_coeffs: Vec<F>,
+}
+
+impl<F: FftField> Config<F> {
+    /// Receive commitments from the transcript: one f̂ commitment per polynomial,
+    /// plus a single batch commitment for all helper polynomials.
+    ///
+    /// Returns `(f_hat_commitments, helper_commitment)`.
+    pub fn receive_commitments<H>(
```
recmo: This should be moved to `committer.rs` to mirror how it's don  in the `whir` impl.

src/protocols/whir_zk/config.rs
recmo: Right now `mod.rs` is basically empty and this file is small. It is also the most public part of the module. We should move the contents of this file to `mod.rs` and delete `config.rs`.

src/algebra/polynomials/mod.rs
```diff
@@ -9,6 +9,37 @@ use ark_ff::Field;
...
+/// Each check is O(n) via `evaluate`, giving O(n) total instead of the O(n log n)
+/// wavelet transform + O(n) allocation that a full comparison would require.
+///
+/// Always returns `true` in release builds (compiles to no-op).
+#[allow(unused_variables)]
+pub fn spot_check_evals_eq<F: Field>(
+    eval_list: &EvaluationsList<F>,
+    coefficients: &CoefficientList<F>,
+) -> bool {
+    #[cfg(debug_assertions)]
```
recmo: This is dangerous and unnecessary. Either mark the whole fn `#[cfg(debug_assertions)]` (so compile fails when you accidentally use it in release), or make it work correctly in release builds as well.

Given the way it's implemented, I don't see a use case in release builds, so I would condition the whole fn. (A more release ready version would use a configurable number of randomly selected indices).

src/algebra/mod.rs
```diff
@@ -119,3 +126,73 @@ pub fn mixed_dot<F: Field, G: Field>(
 
     result
 }
+
+/// Project an extension field element to its base prime field component.
+///
+/// Panics if the element does not lie in the base prime subfield.
+#[inline]
+pub fn project_to_base<F: Field>(val: F) -> F::BasePrimeField {
```
recmo: *Note*: we should never need a function like this. Instead we should keep elements around in their source field.

We also shouldn't assume the embedding is the basefield (even though it is currently hardcoded in whir, this should change).

src/protocols/whir/verifier.rs
```diff
@@ -219,43 +221,69 @@ impl<F: FftField> Config<F> {
...
+            &final_sumcheck_randomness,
+            the_sum,
+        )
+    }
+
+    /// Verify the final consistency check: compute folding randomness, read deferred
+    /// hints, evaluate weight functions, and check the final sumcheck equation.
+    ///
+    /// This is shared between `verify` and `whir_zk::Config::verify`.
+    pub(crate) fn verify_final_consistency<H>(
```
recmo: This one is a bit less clear-cut than the other two fns. It's complex enough to deserve it's own fn instead of duplicating the code.

In any case it is if we use it in both `whir` and `whir_zk` it shouldn't be part of `whir::Config` but be a standalone fn.

recmo: I'm also not against duplicating this code. Odds are good that we want to do things slightly differently in `whir_zk`?

src/protocols/whir_zk/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+/// blinding polynomial WHIR config (for blinding polynomials committed
+/// in a separate Merkle tree).
+///
+/// ZK parameters (`num_witness_variables`, `num_blinding_variables`) are
+/// derivable from the two configs and exposed as methods.
+#[derive(Clone, Serialize)]
+#[serde(bound = "F: FftField")]
+pub struct Config<F: FftField> {
+    /// Main WHIR config for the witness polynomial.
+    pub main: whir::Config<F>,
```
recmo: `main` is not very descriptive. Maybe `blinded_poly_commitment` ?

src/protocols/whir_zk/utils.rs
```diff
@@ -0,0 +1,364 @@
...
+};
+
+// ── IRS Domain Parameters (shared by prover & verifier) ──────────────────
+
+/// Precomputed IRS domain structure for the initial commitment.
+///
+/// Both prover and verifier need the same domain generators, coset roots, and
+/// sub-domain powers to compute gamma query points. This struct deduplicates
+/// that computation.
+pub(crate) struct IrsDomainParams<F: FftField> {
```
recmo: Instead of having this struct we should just add a bunch of member fns to `Config`. Some of these already exist (e.g. `config.main.whir.initial_committer.generator()` should be the same as `omega_full`). You're welcome to add convenience fns for them like you did with `num_variables`.

src/protocols/whir_zk/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+}
+
+/// Witness: contains commitment witnesses for all ZK components.
+///
+/// Produced by [`Config::commit`] and consumed by [`Config::prove`].
+#[derive(Clone)]
+pub struct Witness<F: FftField> {
+    /// Witnesses for [[f̂₁]] = [[f₁ + msk₁]], ..., [[fₙ]] = [[fₙ + mskₙ]] in main WHIR
+    pub f_hat_witnesses: Vec<irs_commit::Witness<F::BasePrimeField, F>>,
```
recmo: We should not make the internals of `Witness` and `Commitment` pub. (I know it's currently the case in some places, but that's a bug and should be fixed). Generally only these are implementation details of the protocol that should not be exposed.

recmo (changes requested):

PR #223, opened 2026-02-16
Use plain vectors and avoid MLE coefficient form.

recmo:
WHIR is a vector commitment protocol that can open on arbitrary linear forms. This PR reinforces that perspective by avoiding unnecessary interpretations as polynomials and unnecessary conversion to coefficient form.

### Transcript Changes

- The RS encoded message is now the plain vector (before it was the coefficient vector).
- Final vector is now also plain vector (before it was the coefficient vector).

### Changes

- [x] Use plain Vec<F> for vector and covector.
- [x] Drop coefficient form
- [x] Avoid coefficient form altogether in prover.
- [x] Change API to take plain vectors.
- [x] Change linear form `evaluate` semantics to be over un-transformed vectors.
- [x] Adapt tests to be purely vector.
- [x] Convert `Evaluate` to pure evals base only.
- [x] Change `UnivariateEvaluation` semantics to be direct.
- [x] Change RS encoding to work on evaluation basis directly (i.e. skip the wavelet transform step).


Cargo.toml
```diff
@@ -65,18 +64,14 @@ divan = { version = "4.2", package = "codspeed-divan-compat" }
...
+# lto = "fat"
 
 [features]
 default = ["parallel"]
 # Verifier panics instead of returning errors. Useful for debuging.
 verifier_panics = []
-parallel = [
-    "dep:rayon",
-    "ark-poly/parallel",
-    "ark-ff/parallel",
```
Bisht13: Was removing `ark-ff/parallel` intentional?

src/algebra/linear_form/multilinear_extension.rs
Bisht13: Nit: MultilinearEvaluation → MultilinearExtension

WizardOfMenlo (approved):
Very happy to see these changes! I really like not thinking about polynomials unless we absolutely have to, and code seems much simpler now.

PR #225, opened 2026-02-18
Ownership-based prove() with buffer recycling and generic Config

Bisht13:
## Summary

- Refactors `prove()` to take owned/Cow arguments, enabling in-place buffer reuse and early deallocation
- Recycles the first input vector as the RLC accumulator via `map_vec` (zero-cost for Identity embedding)
- Recycles Covector buffers via `Any` downcasting to avoid a `vec![F::ZERO; n]` allocation
- Makes `Config` generic over the embedding type (`Config<F, M>` with `M = Basefield<F>` default)
- Adds `shrink_to_fit()` in `sumcheck::fold` to release excess capacity after each round

## Motivation

When used from provekit with BN254 (32-byte field elements), initial witnesses alone consume ~570 MB. The previous borrow-based API kept all caller data alive for the entire prove duration, causing ~970 MB peak usage. With ownership transfer, memory is freed as soon as each structure is consumed, reducing peak memory significantly.

## Changes
- `src/protocols/whir/prover.rs` — Cow-based prove() API, vector/covector recycling, `map_vec` usage
- `src/protocols/whir/config.rs` — `Config<F, M = Basefield<F>>` generic over embedding
- `src/protocols/whir/committer.rs` — Generic `Witness<F, M>`, updated `commit()`
- `src/protocols/whir/verifier.rs` — Generic `verify()`
- `src/algebra/embedding.rs` — `map_vec` on Embedding trait with Identity zero-cost override
- `src/algebra/linear_form/mod.rs` — `Any` supertrait on `LinearForm` for downcasting
- `src/algebra/sumcheck.rs` — `shrink_to_fit()` in `fold`
- `src/protocols/whir/mod.rs` — Test cleanup: `build_prove_forms` helper
- `src/bin/main.rs`, `src/bin/benchmark.rs` — Updated call sites
## Testing
All 141 tests pass, clippy clean, fmt clean.

### DIFF
--- a/src/protocols/whir/prover.rs
+++ b/src/protocols/whir/prover.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{any::Any, borrow::Cow};
 
 use ark_ff::FftField;
 use ark_std::rand::{CryptoRng, RngCore};
@@ -91,7 +91,7 @@ where
             zip_strict(&linear_forms, evaluations.chunks_exact(num_vectors))
         {
             use crate::algebra::linear_form::Covector;
-            let covector = Covector::from(&**linear_form);
+            let covector = Covector::from(linear_form.as_ref());
             for (vector, evaluation) in zip_strict(&vectors, evaluations) {
                 debug_assert_eq!(covector.evaluate(self.embedding(), vector), *evaluation);
             }
@@ -164,7 +164,7 @@ where
                 .zip(linear_forms.iter())
                 .enumerate()
                 .find_map(|(i, (&coeff, form))| {
-                    form.as_any()
+                    (form.as_ref() as &dyn Any)
                         .downcast_ref::<Covector<F>>()
                         .map(|cov| (i, coeff, &cov.vector))
                 });
@@ -196,6 +196,7 @@ where
         .sum();
 
         drop(evaluations);
+
         debug_assert!(!has_constraints || dot(&vector, &covector) == the_sum);
 
         // Add OODS constraints
@@ -206,6 +207,7 @@ where
 
         drop(oods_evals);
         drop(oods_matrix);
+
         debug_assert!(!has_constraints || dot(&vector, &covector) == the_sum);
 
         // Run initial sumcheck on batched vectors with combined statement
--- a/src/algebra/linear_form/covector.rs
+++ b/src/algebra/linear_form/covector.rs
@@ -1,5 +1,3 @@
-use std::any::Any;
-
 use ark_ff::Field;
 
 use super::{Evaluate, LinearForm};
@@ -26,10 +24,6 @@ impl<F: Field> LinearForm<F> for Covector<F> {
 
     fn accumulate(&self, accumulator: &mut [F], scalar: F) {
         scalar_mul_add(accumulator, scalar, &self.vector);
-    }
-
-    fn as_any(&self) -> &dyn Any {
-        self
     }
 }
 
--- a/src/algebra/linear_form/mod.rs
+++ b/src/algebra/linear_form/mod.rs
@@ -55,9 +55,6 @@ pub trait LinearForm<F: Field>: Any {
     ///
     /// This function is only called by the prover.
     fn accumulate(&self, accumulator: &mut [F], scalar: F);
-
-    /// Upcast to `Any` for dynamic downcasting of concrete types.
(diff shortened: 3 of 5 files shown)


recmo:
Hmm... we definitely want this optimization. But I can also imagine there are legit usecases where we don't want the prover to destroy the vecs. (Multiple whir::prove calls using the same commitments?).

Can we use `Vec<Cow<[F]>>` `Vec<Cow<Witness<F>>>`, etc. instead to support both?

src/protocols/whir/prover.rs
```diff
@@ -121,12 +133,21 @@ impl<F: FftField> Config<F> {
...
+            mixed_scalar_mul_add(
+                self.embedding(),
+                &mut vector,
+                *rlc_coeff,
+                input_vector.as_slice(),
+            );
         }
+
+        // ── Input vectors are fully consumed; free them. ──
+        drop(vectors);
```
recmo: Instead of dropping here you can use `.into_iter()`. This is done by simply removing the `&` from `&vectors` in `zip_strict`.

src/protocols/whir/prover.rs
```diff
@@ -179,39 +207,49 @@ impl<F: FftField> Config<F> {
             self.initial_sumcheck
                 .prove(prover_state, &mut vector, &mut covector, &mut the_sum)
         };
+
+        // Reclaim memory from the folded-away upper halves.
+        vector.shrink_to_fit();
+        covector.shrink_to_fit();
```
recmo: Push this in all the way to `algebra::sumcheck::fold`.

There may be a slight performance penalty to this: the allocator may decide to relocate the buffer which causes a memcopy. I would not worry about this until we see evidence of it slowing things down.

src/protocols/whir/prover.rs
```diff
@@ -121,43 +133,80 @@ impl<F: FftField> Config<F> {
...
         // Random linear combination of the vectors.
-        let mut vector_rlc_coeffs: Vec<F> = geometric_challenge(prover_state, vectors.len());
-        let mut vector = vec![F::ZERO; self.initial_size()];
+        let mut vector_rlc_coeffs: Vec<F> = geometric_challenge(prover_state, num_vectors);
         assert_eq!(vector_rlc_coeffs[0], F::ONE);
-        for (rlc_coeff, &input_vector) in zip_strict(&vector_rlc_coeffs, vectors) {
-            mixed_scalar_mul_add(self.embedding(), &mut vector, *rlc_coeff, input_vector);
+        // Recycle the first input as the accumulator (its coefficient is always ONE).
+        let mut vectors = vectors.into_iter();
+        let mut vector = lift(self.embedding(), &vectors.next().expect("non-empty"));
```
recmo: this takes a reference so doesn't actually recycle anything (it allocates a new vec in lift fn). We also can't recycle unless `M = Identity<_>`.

I don't think we can solve this without specialization. One way is to add a map_vec(&self, vec: Vec<M::Source>) -> Vec<M::Target> fn to the Embedding trait. We can give it a default implementation that does vec.into_par_iter().map(|e| self.map(e)).collect()  and then override in Identity to just return vec. (edited)

src/protocols/whir/prover.rs
```diff
@@ -121,43 +135,78 @@ impl<F: FftField> Config<F> {
...
         UnivariateEvaluation::accumulate_many(&oods_evals, &mut covector, oods_rlc_coeffs);
-        the_sum += zip_strict(oods_rlc_coeffs, oods_matrix.chunks_exact(vectors.len()))
+        the_sum += zip_strict(oods_rlc_coeffs, oods_matrix.chunks_exact(num_vectors))
             .map(|(poly_coeff, row)| *poly_coeff * dot(&vector_rlc_coeffs, row))
             .sum::<F>();
 
-        // These invariants are maintained throughout the proof.
-        debug_assert_eq!(dot(&vector, &covector), the_sum);
+        drop(oods_evals);
+        drop(oods_matrix);
```
recmo: Conceptually for me these are 'cleanup after last use' and as such they belong to the code block before. But now they are grouped with the debug invariant check, which feels wrong.

-> move the empty line between drop and debug_assert.

recmo (approved):
A simplification and two minor code style issues. After that LGTM!

PR #226, opened 2026-02-18
Feat : added WHIR protocol 2.0 implementation

ocdbytes:
- Previously closed PRs (#216 #219 #222 ) were related to this. but due to optimisation pushes in main branch had to create this new PR based on latest changes. 
- De duplicated implementation of zk whir 2. All the critical code now lives in whir protocol crate.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -73,27 +73,47 @@ fn main() {
         args.pow_bits = Some(default_max_pow(args.num_variables, args.rate));
     }
 
+    if let Err(err) = validate_args(&args) {
+        eprintln!("Error: {err}");
+        std::process::exit(2);
+    }
+
     runner(&args, field);
 }
 
-fn runner(args: &Args, field: AvailableFields) {
-    if args.zk {
-        // zkWHIR is currently pinned to identity embedding over a single field.
-        match field {
-            AvailableFields::Goldilocks1 => run_whir_pcs_zk::<fields::Field64>(args),
-            AvailableFields::Field128 => run_whir_pcs_zk::<fields::Field128>(args),
-            AvailableFields::Field192 => run_whir_pcs_zk::<fields::Field192>(args),
-            AvailableFields::Field256 => run_whir_pcs_zk::<fields::Field256>(args),
-            AvailableFields::Goldilocks2 | AvailableFields::Goldilocks3 => {
-                println!(
-                    "Error: --zk currently supports only single-field configurations (Identity embedding)."
-                );
-            }
-        }
-        return;
+fn validate_args(args: &Args) -> Result<(), String> {
+    if args.zk && !matches!(args.protocol_type, WhirType::PCS) {
+        return Err("--zk is only supported with --type PCS".into());
     }
+    Ok(())
+}
 
-    // Type reflection on field
+fn runner(args: &Args, field: AvailableFields) {
+    match (args.protocol_type, args.zk, field) {
+        // zkWHIR is currently pinned to identity embedding over a single field.
+        (WhirType::PCS, true, AvailableFields::Goldilocks1) => {
+            run_whir_pcs_zk::<fields::Field64>(args);
+        }
+        (WhirType::PCS, true, AvailableFields::Field128) => {
+            run_whir_pcs_zk::<fields::Field128>(args);
+        }
+        (WhirType::PCS, true, AvailableFields::Field192) => {
+            run_whir_pcs_zk::<fields::Field192>(args);
+        }
+        (WhirType::PCS, true, AvailableFields::Field256) => {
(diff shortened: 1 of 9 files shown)


src/protocols/whir_zk/committer.rs
```diff
@@ -0,0 +1,144 @@
+use ark_ff::FftField;
+use ark_std::rand::{rngs::StdRng, SeedableRng};
+
+use super::{utils::BlindingPolynomials, Config};
+use crate::{hash::Hash, transcript::ProverMessage};
+
+/// zkWHIR commitment object.
+pub struct Commitment<F: FftField> {
+    pub f_hat: Vec<crate::protocols::whir::Commitment<F>>,
+    pub blinding: crate::protocols::whir::Commitment<F>,
```
recmo: We generally don't use full paths to not clutter the code. The protocols are a little special because they repeat a lot of struct names (Config, Witness, etc.) In this case we should `use crate::protocols::whir` and then use `whir::Commitment`. Same for the other protocols.

src/protocols/whir_zk/verifier.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn verify<H>(
+        &self,
+        verifier_state: &mut crate::transcript::VerifierState<'_, H>,
+        weights: &[&dyn crate::algebra::linear_form::LinearForm<F>],
+        evaluations: &[F],
+        commitment: &super::Commitment<F>,
+    ) -> crate::transcript::VerificationResult<()>
+    where
+        H: crate::transcript::DuplexSpongeInterface,
+        F: crate::transcript::Codec<[H::U]>,
```
recmo: Use `use` instead of full paths.

src/protocols/whir_zk/verifier.rs
```diff
@@ -0,0 +1,137 @@
...
+        [u8; 32]: crate::transcript::Decoding<[H::U]>,
+        crate::transcript::codecs::U64: crate::transcript::Codec<[H::U]>,
+        Hash: ProverMessage<[H::U]>,
+    {
+        assert_eq!(
+            self.blinded_commitment.initial_committer.num_vectors, 1,
+            "zkWHIR currently expects one vector per commitment"
+        );
+        let num_polynomials = commitment.f_hat.len();
+        if evaluations.len() != weights.len() * num_polynomials {
```
recmo: Prefer the `verify!` macro which can conditionally panic. Panicking gives a proper backtrace and makes debugging a lot eaier.

recmo (changes requested):
Some quick code-style issues. Will follow with more in-depth review.

src/protocols/whir/config.rs
```diff
@@ -267,25 +266,20 @@ where
...
-            return false;
-        }
-        for round_config in &self.round_configs {
-            if round_config.pow.difficulty() > max_bits {
-                return false;
-            }
-            if round_config.sumcheck.round_pow.difficulty() > max_bits {
-                return false;
-            }
+    /// Convert a basefield-embedded WHIR config into a single-field Identity embedding.
```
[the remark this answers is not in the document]
recmo: It's also not necessarily sounds as the whir configuration is only valid for a particular set of field sizes. I think going from basefield to identity makes soundness strictly stronger, but parameters will be suboptimal.

src/transcript/mod.rs
```diff
@@ -20,6 +21,51 @@ pub use spongefish::{
...
 };
 
+/// Max bytes for stack-encoding a field element (supports up to Fp12 over 64-byte base fields).
+const MAX_FIELD_BYTES: usize = 768;
+
+/// Encode a field element into `buf` without heap allocation.
+/// Matches spongefish's `Encoding<[u8]>` for arkworks field elements.
+/// Returns the number of bytes written.
+#[inline]
+fn encode_field_into<F: Field>(elem: &F, buf: &mut [u8]) -> usize {
```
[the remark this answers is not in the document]
ocdbytes: This is for the `prover_message_field` function. Are we reverting that change to use `prover_message` ? I thought we agreed on keeping this for reduced number of allocations.

src/protocols/whir/config.rs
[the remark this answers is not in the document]
ocdbytes: removed to_identity function and defaulting to basefield without any user selection for this PR. We can create an issue to support extension field based on user's choice.

src/bin/main.rs
```diff
@@ -94,6 +113,11 @@ where
             run_whir_pcs::<F>(args);
         }
         WhirType::LDT => {
+            if args.zk {
```
Bisht13: This if args.zk check is unreachable. When args.zk is true, runner() handles it and returns early (line 90) before run_whir() is ever called.

PR #227, opened 2026-02-20
Refactor protocol configuration.

recmo:
- [x] Simplify `WhirParameters` and `whir::Config`.
- [x] Fix missing factor 7 in mca_error.
- [x] Fix missing η in Johnson query error.
- [x] Fix incorrect sign in Unique decoding query error.
- [x] Compute using fractional field size.
- [x] Push inward into the protocols.

### Later 

- Implement latest MCA bounds.
- Per round PoW in sumcheck.

### DIFF
--- a/benches/expand_from_coeff.rs
+++ b/benches/expand_from_coeff.rs
@@ -33,7 +33,7 @@ fn interleaved_rs_encode(bencher: Bencher, case: &(u32, usize, usize)) {
         .bench_values(|(coeffs, expansion, coset_sz)| {
             black_box(ntt::interleaved_rs_encode(
                 &[&coeffs],
-                coeffs.len() * expansion,
+                (coeffs.len() >> coset_sz) * expansion,
                 1 << coset_sz,
             ))
         });


src/protocols/whir/config.rs
```diff
@@ -1,468 +1,246 @@
...
-                    log_inv_rate as f64,
-                    log_eta_start,
-                ) + (whir_parameters.folding_factor.at_round(0) as f64)
-                    .log2();
-                (whir_parameters.security_level as f64 - prox_gaps_error).max(0.0)
-            }
+        // If we skip the initial sumcheck, we do this pow instead:
+        let initial_skip_pow_bits = {
+            let prox_gaps_error = initial_committer.rbr_soundness_fold_prox_gaps()
+                + (whir_parameters.initial_folding_factor as f64).log2();
```
WizardOfMenlo: Re my later comment: The prox gaps error is not constant across sumcheck rounds (even in a single iteration). See for example lemma 6.6 in https://eprint.iacr.org/2025/753.pdf (the e_sum,i term). We can save some pow this way.

Issue #230, opened 2026-03-08
Opt : ZK WHIR with more optimised parameters for shorter proof sizes


ocdbytes:
Reference Doc : [[ ZK WHIR updated params doc by @yswami-tfh ]](https://www.notion.so/worldcoin/Zero-Knowledge-WHIR-with-poly-logarithmic-overhead-2e28614bdf8c80be8c95edaa2fd93f9f?source=copy_link#3148614bdf8c80faa8d5e69b24b833b6)

Changes Overview : 
- New Φᵢ morphisms and basis change logic
- Explicit OOD/STIR handling in the ZK wrapper, and fold_args computation
- Support for ⌊μ/ℓ⌋+1 polynomials instead of μ+1

Refactoring and logical Change :
- Need to find a way to intercept the queries in whir instances created in zk WHIR. Right now It is being handled in the wrapper and we calculate gamma coset points in wrapper and evaluate them and then send them to blinding whir instance which handles it own its own.

PR #239, opened 2026-03-23
Opt : smaller proof size by using alternate randomness sampling approach (audit branch : v1)

ocdbytes:
**Summary**

- Implements zkWHIR 2.0 using the "Alternative Randomness Sampling" approach, replacing the previous per-round blinding strategy. Proof size drops from (μ+1)·q(δ) to (ν+1)·q(δ) field
  elements, where ν = ⌊μ/ℓ⌋ + 1 ≪ μ.
- Extracts shared WHIR round logic into whir/rounds.rs, deduplicating the fold-commit-sumcheck loop between base WHIR and zkWHIR.
- Adds open_at_indices / verify_at_indices to irs_commit for the Γ consistency check (opening at caller-provided codeword positions instead of transcript-sampled ones).

**Design**

Two WHIR instances run as sub-protocols:
1. Blinded polynomial (f̂ = f + msk(Φ₀)): standard WHIR rounds on f_zk = ρ·f + g, with a Γ consistency check verifying [[f̂]] openings match [[H]].
2. Blinding polynomial (M, ĝ₁..ĝ_ν): batched proof over ν + n committed vectors using weight covectors derived from beq tables.

**Protocol steps:**

1. Commitment — sample masking/blinding polynomials, commit both instances
2. Blinding claims — verifier sends β, prover sends G claims
3. Combination — verifier sends ρ ≠ 0, prover forms f_zk = ρ·f + g
4. Initial sumcheck on f_zk
5. OOD/STIR queries + remaining WHIR rounds (shared via rounds.rs)
6. Γ consistency check — verify [[f̂]] openings match [[H]] at FRI query indices
7. Batched blinding proof via second WHIR instance

Reference Issue : #230 

Reference Doc : [[ ZK WHIR updated params doc by @yswami-tfh ]](https://www.notion.so/worldcoin/Zero-Knowledge-WHIR-with-poly-logarithmic-overhead-2e28614bdf8c80be8c95edaa2fd93f9f?source=copy_link#3148614bdf8c80faa8d5e69b24b833b6)

| Metric                | Previous Implementation | New Implementation | Change |
|----------------------|------------------------|--------------------|--------|
| Proof Size           | 3.2 MB                 | 740 KB             | -77%   |
| Verifier Time        | 243 ms                 | 136 ms             | -44%   |


Bisht13 (changes requested):

src/protocols/whir_zk/verifier.rs
```diff
@@ -1,49 +1,489 @@
...
+            // SECURITY
+            // Consumed for Fiat-Shamir transcript binding only — NOT directly
+            // verified. The prover sends MLE(f̂_combined, fold_args(r̄, z)),
+            // which differs from the univariate fold evaluation fold(r̄, f̂)(z).
+            // Soundness: the STIR constraint accumulation (Step 5f) binds [[H]]
+            // evaluations via the sumcheck chain, and the Γ consistency check
+            // (Step 6) verifies fold(r̄, [[f̂]])(γ) at FRI query indices.
+            // Together these ensure the prover cannot forge the fold without
+            // detection.
+            let _ood_f_hat: F = self.verifier_state.prover_message()?;
```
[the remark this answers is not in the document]
ocdbytes: according to current doc spec and the protocol implementation verifier cannot construct is locally because verifier doesn't have f_hat poly to construct MLE here.

src/protocols/whir_zk/mod.rs
```diff
@@ -1,240 +1,274 @@
...
+        // will underflow when computing m_cap. Catch misconfiguration here with a
+        // clear message instead of a confusing panic deep in the proving path.
+        assert!(
+            ell >= params.initial_folding_factor,
+            "ell={ell} must be >= initial_folding_factor={} \
+             (parameters too aggressive for ZK sizing)",
+            params.initial_folding_factor
         );
-        num_blinding_variables
-    }
```
Bisht13, line 152: zkWHIR indexes round_configs[0] directly in 4 places (prover:328, verifier:212, utils:458, utils:462) without guarding for empty. Base WHIR handles the 0-round case, but zkWHIR would panic. The ell-sizing asserts make this practically unreachable, but an explicit guard would fail clearly:
```rs
assert!(
    !blinded_config.round_configs.is_empty(),
    "zkWHIR 2.0 requires at least one WHIR round \
     (num_variables_main too small for folding_factor)"
);
```

src/protocols/whir_zk/verifier.rs
```diff
@@ -1,49 +1,503 @@
...
+        // verification, matching the base WHIR pattern.
+        // =====================================================================
+        let evaluation_point: Vec<F> = round_folding_randomness
+            .into_iter()
+            .flat_map(|p| p.0.into_iter())
+            .collect();
+
+        let poly_eval = MultilinearExtension::new(remaining.final_sumcheck_randomness.0)
+            .evaluate(&Identity::new(), &remaining.final_vector);
+        let mut linear_form_rlc = prepare.the_sum / poly_eval;
```
Bisht13, line 373: prepare.the_sum / poly_eval, no guard against poly_eval == 0. Field division panics on zero denominator. A verify!(poly_eval != F::ZERO) before the division would be consistent with the existing verify!(rho != F::ZERO) pattern and turn a panic into a clean verifier rejection. (Note: base WHIR has the same pattern, could be addressed repo-wide.)

Bisht13 (changes requested):

PR #242, opened 2026-03-26
Drop in SmallFp Goldilocks

z-tech:
## What does this PR do?

- Just drops in the newly merged [SmallFp impl](https://github.com/arkworks-rs/algebra/pull/1044) of the Goldilocks field
- I don't exactly know how these benches work but the ones I found show some improvement
- Presumably there are extended benches that could be run to see more results

## Results of running cargo bench
`SmallFp` vs `Fp64<MontBackend>` for the Goldilocks field ($p = 2^{64} - 2^{32} + 1$). Same machine, single-threaded, `--release`.
### Interleaved RS Encode (median)
| Size | `Fp64` | `SmallFp` | Δ |
|------|:---:|:---:|:---:|
| (16, 2, 2) | 1.241 ms | 1.053 ms | **−15%** |
| (16, 4, 3) | 2.450 ms | 1.909 ms | **−22%** |
| (18, 2, 2) | 4.579 ms | 4.521 ms | −1% |
| (18, 4, 3) | 7.857 ms | 8.662 ms | +10% |
| (20, 2, 3) | 17.17 ms | 16.18 ms | **−6%** |
| (20, 4, 4) | 35.58 ms | 33.92 ms | **−5%** |
| (22, 4, 4) | 163.5 ms | 160.5 ms | **−2%** |
### Sumcheck First Round (median)
| Size | `Fp64` | `SmallFp` | Δ |
|------|:---:|:---:|:---:|
| 65536 | 322.9 µs | 230.7 µs | **−29%** |
| 262144 | 577.1 µs | 460.9 µs | **−20%** |
| 1048576 | 1.490 ms | 1.228 ms | **−18%** |


src/algebra/fields.rs
```diff
@@ -80,24 +82,42 @@ pub type Field192 = Fp192<MontBackend<FConfig192, 3>>;
...
+pub type Field64 = SmallFp<FConfig64>;
+
+/// Compute the Montgomery representation of a Goldilocks field element at
+/// compile time.
+///
+/// For SmallFp<u64> the Montgomery constant R = 2^k where k = ceil(log2(p)).
+/// Since p ≈ 2^64, we have k = 64 and R = 2^64.
+///
+/// Montgomery form of `v` is `v · R mod p`, computed in u128 to avoid overflow.
+const fn goldilocks_mont(v: u64) -> u64 {
```
recmo: Why is this not part of `SmallFp`? I would expect an

```
impl SmallFp {
   pub const fn from_u64(n: u64) -> Self;
}
```

or similar being generated by the macro that does exactly this.

It's also leaks the abstraction in that it assumes `SmallFp` uses Montogomery representation, which is not obvious for small field (i.e. M31 works better without IIRC).

recmo (changes requested):

src/algebra/fields.rs
z-tech: Fair point wrote the little helper here: https://github.com/arkworks-rs/algebra/pull/1084

Also, we're happy to further optimize any Mersenne primes when/ if you have a specific need: https://andrewzitek.xyz/smallfp-site/

I'm confident this would be a reasonable change in the framework we've created.

Cargo.toml
```diff
@@ -101,3 +101,8 @@ lto = "thin" # LTO breaks line attribution.
 # Enable a large amount of optimization in the dev profile for dependencies.
 [profile.dev.package."*"]
 opt-level = 3
+
+[patch.crates-io]
+ark-ff = { git = "https://github.com/arkworks-rs/algebra" }
+ark-std = { git = "https://github.com/arkworks-rs/std" }
+ark-serialize = { git = "https://github.com/arkworks-rs/algebra" }
```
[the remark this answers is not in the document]
z-tech: This merged into Arkworks 0.6.0 btw

PR #249, opened 2026-04-03
zkWHIR 3.0 : Code Switch Protocol (part 1)

ocdbytes:
**Summary**

Implements the code-switching IOR from Section 9.4 as a standalone sub-protocol in src/protocols/code_switch.rs.
Code switching reduces a proximity claim about oracle f w.r.t. source code C to a proximity claim about oracle g w.r.t. target code C', enabling the polylogarithmic-verifier protocol (Theorem 4) where the tested code shrinks each round.

**What's implemented**

  - Config: Derives target code C' from source (ℓ' = ℓ, m' from rate), optional ZK mask code C_zk, OOD sample computation from Lemma 9.9. ZK query budgets passed via Option<ZkQueryBudget>.
  - Prover (prove): Commits target oracle g, optionally commits mask oracle s = Enc_{C_zk}((r, pad), r''), sends OOD answers, opens source oracle. Reuses irs_commit::commit and irs_commit::open with zero duplicated logic.
  - Verifier (verify): Mirrors prover transcript, computes batched target μ', builds constraint weights for output sl'. Mask weights have pre-shifted RLC coefficients via MaskClaimInfo.
  - Shared batching_challenge helper ensures prover/verifier transcript sync.
  - Shared num_ood_samples extracted to irs_commit — used by both irs_commit::Config::new and code switch.
  - Tests: Proptest over field types, sizes, rates, and ZK/non-ZK mode. Tests here just check the code switch transcript match. (Tests will be added in future PRs).

  **What's NOT in this PR (next PRs)**

  - Integration into whir's round structure as an alternative to sumcheck-only folding
  - Implement zk sumcheck according to the construction 6.3 to include masks
  - Refactoring irs_commit to separate commitment from OOD (deferred to whir integration)
  - Parameter selection
  - Mask resolution per iteration (this makes the basecase verification simple and whole flow clean and contained)

### DIFF
--- a/src/protocols/code_switch.rs
+++ b/src/protocols/code_switch.rs
@@ -1,0 +1,432 @@
+//! Code-switching IOR: R_{C, sl} → R_{C', sl'}
+//!
+//! Reduces a proximity claim about oracle f (source code C) to a proximity
+//! claim about oracle g (target code C'). Non-ZK variant.
+//!
+//! Paper: Construction 9.7 (p.55), Theorem 9.6 (p.54), Lemma 9.9 (p.57)
+
+use ark_ff::Field;
+use ark_std::rand::{distributions::Standard, prelude::Distribution, CryptoRng, RngCore};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    algebra::{
+        dot,
+        embedding::{Embedding, Identity},
+        fields::FieldWithSize,
+        linear_form::UnivariateEvaluation,
+        mixed_dot, mixed_univariate_evaluate, univariate_evaluate,
+    },
+    bits::Bits,
+    hash::Hash,
+    protocols::{geometric_challenge::geometric_challenge, irs_commit, proof_of_work},
+    transcript::{
+        Codec, Decoding, DuplexSpongeInterface, ProverMessage, ProverState, VerificationResult,
+        VerifierMessage, VerifierState,
+    },
+    type_info::Typed,
+};
+
+/// Prover output from the code-switch interaction phase.
+/// Paper: Equation 9 (p.55)
+#[derive(Clone, Debug)]
+pub struct Witness<F: Field, G: Field = F> {
+    /// Target oracle commitment data. G_target = F since target uses Identity<F>.
+    pub target_witness: irs_commit::Witness<F>,
+    /// The message f, taken by ownership from the caller.
+    pub message: Vec<F>,
+    /// Source oracle evaluations at in-domain query points (in M::Source = G).
+    pub source_evaluations: irs_commit::Evaluations<G>,
+}
+
+/// Verifier output from the code-switch decision phase.
+/// Paper: Equation 9 (p.55) — output (x', y')
+#[derive(Clone, Debug)]
+pub struct CodeSwitchClaim<F: Field, G: Field = F> {
+    /// μ' — batched target value. Decision phase formula (p.55).
+    pub mu_prime: F,
+    /// ν_1 = batching_coeffs[0]. Caller scales original sl by this.
+    pub original_sl_coeff: F,
+    /// Target oracle commitment (g).
+    pub target_commitment: irs_commit::Commitment<F>,
+    /// RLC coefficients for constraint weights (batching_coeffs[1..]).
+    pub constraint_rlc_coeffs: Vec<F>,
+    /// OOD constraint weights — UnivariateEvaluation at each OOD point (M::Target).
+    pub ood_constraint_weights: Vec<UnivariateEvaluation<F>>,
+    /// In-domain constraint weights — UnivariateEvaluation at each source eval point (M::Source).
+    pub in_domain_constraint_weights: Vec<UnivariateEvaluation<G>>,
+    /// Source oracle evaluation data from in-domain queries.
+    pub source_evaluations: irs_commit::Evaluations<G>,
+}
+
+/// Code-switching IOR config. Non-ZK variant (n=0, no C_zk masks).
+///
+/// TODO [ZK]: Add mask oracle, private zero-evaders, mask terms in sl'.
+#[derive(Clone, PartialEq, Eq, Debug, Hash, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct Config<M: Embedding> {
+    pub embedding: Typed<M>,
+    /// Source code C.
+    pub source: irs_commit::Config<M>,
+    /// Target code C'. Message length ℓ' = ℓ (same message).
+    pub target: irs_commit::Config<Identity<M::Target>>,
+    /// OOD samples tying source and target to the same message.
+    pub cs_ood_samples: usize,
+    // In-domain queries reuse source.in_domain_samples (same formula).
+    // Reintroduce cs_in_domain_queries if security budgets diverge or mask_length > 0.
+    /// Proof-of-work config (set by parent orchestrator).
+    pub pow: proof_of_work::Config,
+}
+
+/// Squeeze batching randomness from the transcript.
+/// Both prove() and verify() must call this at the same transcript position
+/// to stay in sync. Returns (ν_1, batching_coeffs) where ν_1 scales the
+/// original sl and batching_coeffs[i] scales the i-th constraint weight.
+fn batching_challenge<T, F>(transcript: &mut T, count: usize) -> (F, Vec<F>)
+where
+    T: VerifierMessage,
+    F: Field + Decoding<[T::U]>,
+{
+    let mut coeffs = geometric_challenge(transcript, count);
+    let original_sl_coeff = if coeffs.is_empty() {
+        F::ONE
+    } else {
+        coeffs.remove(0)
+    };
+    (original_sl_coeff, coeffs)
+}
+
+impl<M: Embedding> Config<M> {
+    pub fn new(
+        security_target: f64,
+        unique_decoding: bool,
+        hash_id: crate::engines::EngineId,
+        source_config: irs_commit::Config<M>,
+        target_log_inv_rate: usize,
+        target_interleaving_depth: usize,
+    ) -> Self
+    where
+        M: Default,
+        M::Target: Default,
+    {
+        let source_message_length = source_config.message_length();
+
+        // Target: ℓ' = ℓ, m' = ⌈ℓ'/ρ'⌉
+        let target_message_length = source_message_length;
+        let target_rate = 0.5_f64.powf(target_log_inv_rate as f64);
+        let target_vector_size = target_message_length * target_interleaving_depth;
+        let target_config = irs_commit::Config::<Identity<M::Target>>::new(
+            security_target,
+            unique_decoding,
+            hash_id,
+            1,
+            target_vector_size,
+            target_interleaving_depth,
+            target_rate,
+        );
+
+        // OOD: solve |Λ(C',δ')|²/2 · ((ℓ-1)/|F|)^{t_ood} ≤ 2^{-λ}
+        // Target list size, source message length (Lemma 9.9 Error 1, p.57)
+        let cs_ood_samples = irs_commit::num_ood_samples(
+            unique_decoding,
+            security_target,
+            M::Target::field_size_bits(),
+            target_config.list_size(),
+            source_message_length,
+        );
+
+        Self {
+            embedding: Typed::<M>::default(),
+            source: source_config,
+            target: target_config,
+            cs_ood_samples,
+            pow: proof_of_work::Config {
+                hash_id,
+                threshold: proof_of_work::threshold(Bits::new(0.0)),
+            },
+        }
+    }
+
+    /// Prover interaction phase — Construction 9.7 Steps 1-4 (p.55).
+    pub fn prove<H, R>(
+        &self,
+        prover_state: &mut ProverState<H, R>,
+        message: Vec<M::Target>,
+        source_randomness: &[M::Source],
+        source_witness: &irs_commit::Witness<M::Source, M::Target>,
+    ) -> Witness<M::Target, M::Source>
+    where
+        H: DuplexSpongeInterface,
+        R: RngCore + CryptoRng,
+        Standard: Distribution<M::Target>,
+        M::Target: Codec<[H::U]>,
+        u8: Decoding<[H::U]>,
+        Hash: ProverMessage<[H::U]>,
+    {
+        // Step 1: g := Enc_{C'}(f, r')
+        let target_witness = self.target.commit(prover_state, &[&message]);
+
+        // Step 2: receive OOD challenge ρ_ood
+        let ood_points: Vec<M::Target> = prover_state.verifier_message_vec(self.cs_ood_samples);
+
+        // Step 3: send OOD answers y_i = f̂(ρ_i) + ρ_i^ℓ · r̂(ρ_i)
+        let msg_len = message.len();
+        for &point in &ood_points {
+            let f_eval = univariate_evaluate(&message, point);
+            let r_eval =
+                mixed_univariate_evaluate(self.source.embedding(), source_randomness, point);
+            let shift = point.pow([msg_len as u64]);
+            prover_state.prover_message(&(f_eval + shift * r_eval));
+            // TODO [ZK]: add ρ^{ℓ+r} · ŝ(ρ) term for mask oracle
+        }
+
+        // Step 4: open source oracle at in-domain queries
+        let source_evaluations = self.source.open(prover_state, &[source_witness]);
+
+        // Step 5: batching randomness (must match verify's batching_challenge call)
+        let _batching = batching_challenge::<_, M::Target>(
+            prover_state,
+            1 + ood_points.len() + source_evaluations.matrix.len(),
+        );
+
+        Witness {
+            target_witness,
+            message,
+            source_evaluations,
+        }
+    }
+
+    /// Verifier decision phase — Construction 9.7 Steps 1-4 + Decision (p.55).
+    pub fn verify<H>(
+        &self,
+        verifier_state: &mut VerifierState<H>,
+        mu: M::Target,
+        source_commitment: &irs_commit::Commitment<M::Target>,
+    ) -> VerificationResult<CodeSwitchClaim<M::Target, M::Source>>
+    where
+        H: DuplexSpongeInterface,
+        Standard: Distribution<M::Target>,
+        M::Target: Codec<[H::U]>,
+        u8: Decoding<[H::U]>,
+        Hash: ProverMessage<[H::U]>,
+    {
+        // Step 1: receive target commitment
+        let target_commitment = self.target.receive_commitment(verifier_state)?;
+
+        // Step 2-3: OOD challenge + answers
+        let ood_points: Vec<M::Target> = verifier_state.verifier_message_vec(self.cs_ood_samples);
+        let ood_answers: Vec<M::Target> =
+            verifier_state.prover_messages_vec(self.cs_ood_samples)?;
+
+        // Step 4: verify source oracle openings
+        let source_evaluations = self.source.verify(verifier_state, &[source_commitment])?;
+
+        // Step 5: batching coefficients (shared helper ensures transcript sync with prover)
+        let t_ood = ood_points.len();
+        let t_times_iota = source_evaluations.matrix.len();
+        let (original_sl_coeff, constraint_rlc_coeffs) =
+            batching_challenge(verifier_state, 1 + t_ood + t_times_iota);
+
+        // Step 6: μ' = ν_1·μ + Σ ν_{1+i}·y_i + ΣΣ ν_{...}·f(x_i)_l
+        let mu_prime = original_sl_coeff * mu
+            + dot(&constraint_rlc_coeffs[..t_ood], &ood_answers)
+            + mixed_dot(
+                self.source.embedding(),
+                &constraint_rlc_coeffs[t_ood..],
+                &source_evaluations.matrix,
+            );
+
+        // Step 7: constraint weights for sl'
+        let source_msg_len = self.source.message_length();
+
+        let ood_constraint_weights: Vec<UnivariateEvaluation<M::Target>> = ood_points
+            .iter()
+            .map(|&point| UnivariateEvaluation::new(point, source_msg_len))
+            .collect();
+
+        let in_domain_constraint_weights: Vec<UnivariateEvaluation<M::Source>> =
+            source_evaluations.evaluators(source_msg_len).collect();
+
+        Ok(CodeSwitchClaim {
+            mu_prime,
+            original_sl_coeff,
+            target_commitment,
+            constraint_rlc_coeffs,
+            ood_constraint_weights,
+            in_domain_constraint_weights,
+            source_evaluations,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use ark_std::rand::{
+        distributions::Standard, prelude::Distribution, rngs::StdRng, Rng, SeedableRng,
+    };
+    use proptest::{proptest, sample::select};
+
+    use super::*;
+    use crate::{
+        algebra::{embedding::Identity, fields, linear_form::Evaluate, ntt, random_vector},
+        transcript::{codecs::U64, DomainSeparator},
+    };
+
+    /// Run full code switch (prove + verify) and check μ' completeness.
+    /// Uses ι=1 (num_vectors=1, interleaving_depth=1) so weight count matches RLC count.
+    fn test_completeness<F: Field + FieldWithSize + Codec<[u8]> + 'static>(
+        seed: u64,
+        source_vector_size: usize,
+        source_log_inv_rate: usize,
+        target_log_inv_rate: usize,
+    ) where
+        Standard: Distribution<F>,
+        Hash: ProverMessage<[u8]>,
+    {
+        let security_target = 32.0;
+        let unique_decoding = false;
+        #[allow(clippy::cast_possible_wrap)]
+        let source_rate = 0.5_f64.powi(source_log_inv_rate as i32);
+
+        let source_config = irs_commit::Config::<Identity<F>>::new(
+            security_target,
+            unique_decoding,
+            crate::hash::BLAKE3,
+            1,
+            source_vector_size,
+            1,
+            source_rate,
+        );
+        let cs_config = Config::<Identity<F>>::new(
+            security_target,
+            unique_decoding,
+            crate::hash::BLAKE3,
+            source_config.clone(),
+            target_log_inv_rate,
+            1,
+        );
+
+        let mut rng = StdRng::seed_from_u64(seed);
+        let message: Vec<F> = random_vector(&mut rng, source_config.message_length());
+        let mu: F = rng.gen();
+
+        let instance = U64(seed);
+        let ds = DomainSeparator::protocol(&cs_config)
+            .session(&String::from("code_switch_proptest"))
+            .instance(&instance);
+
+        // Prover
+        let mut prover_state = ProverState::new_std(&ds);
+        let source_witness = source_config.commit(&mut prover_state, &[&message]);
+        let cs_witness = cs_config.prove(
+            &mut prover_state,
+            message.clone(),
+            &source_witness.masks,
+            &source_witness,
+        );
+        let proof = prover_state.proof();
+
+        // Verifier
+        let mut verifier_state = VerifierState::new_std(&ds, &proof);
+        let source_commitment = source_config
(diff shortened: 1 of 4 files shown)


src/protocols/code_switch.rs
```diff
@@ -0,0 +1,576 @@
...
+    pub mask_commit: Option<irs_commit::Config<Identity<M::Target>>>,
+}
+
+/// Split geometric challenge into (ν_1, rest). Requires count ≥ 1.
+fn batching_challenge<T, F>(transcript: &mut T, count: usize) -> (F, Vec<F>)
+where
+    T: VerifierMessage,
+    F: Field + Decoding<[T::U]>,
+{
+    debug_assert!(count > 0, "batching requires at least one coefficient");
```
Bisht13: debug_assert should be assert, irs_commit uses assert! for equivalent precondition checks (irs_commit.rs:305-311). Crypto protocol invariants must be checked in release builds.

Same applies to lines 225-229: debug_assert_eq → assert_eq.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,576 @@
...
+            1,
+            target_vector_size,
+            target_interleaving_depth,
+            target_rate,
+        );
+
+        let mask_commit = zk.map(|budget| {
+            assert!(budget.target > 0, "ZK requires nonzero target query budget");
+            assert!(budget.mask > 0, "ZK requires nonzero mask query budget");
+            target_config.mask_length = budget.target;
```
Bisht13, line 126: mask_length is mutated after Config::new, which derived johnson_slack, in_domain_samples, and out_domain_samples for rate WITHOUT mask. The exported cfg.target has stale security parameters.

Within this PR it's conservative (list_size is overestimated → more OOD samples). But downstream code using cfg.target to open/verify will have insufficient in_domain_samples for the actual higher rate.

```rs
// NOTE: johnson_slack, in_domain_samples, out_domain_samples were
// derived for rate without mask. Conservative for cs_ood_samples.
// Target config must be reparameterized before downstream use.
```

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,576 @@
...
+            let r_len = self.source.mask_length * self.source.num_messages();
+            let weights = ConstraintWeights {
+                ood: ood_points
+                    .iter()
+                    .map(|&point| UnivariateEvaluation::new(point, mask_config.message_length()))
+                    .collect(),
+                in_domain: source_evaluations.evaluators(r_len).collect(),
+            };
+            let embedding = self.source.embedding();
+            let mut rlc_coeffs = Vec::with_capacity(num_ood + weights.in_domain.len());
```
Bisht13: Manual indexing into constraint_rlc_coeffs[i] and [num_ood + j]. Prefer zip:
```rs
let rlc_coeffs: Vec<F> = constraint_rlc_coeffs[..num_ood]
    .iter()
    .zip(&weights.ood)
    .map(|(&c, w)| c * w.point.pow([source_msg_len as u64]))
    .chain(
        constraint_rlc_coeffs[num_ood..]
            .iter()
            .zip(&weights.in_domain)
            .map(|(&c, w)| c * embedding.map(w.point.pow([source_msg_len as u64])))
    )
    .collect();
```

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,576 @@
...
+        TestResult {
+            claim,
+            message: msg,
+            mu,
+            mask_message: cw.mask_message,
+        }
+    }
+
+    /// Completeness: μ' = ν_1·μ + ⟨f, source_weights⟩ + ⟨mask_msg, mask_weights⟩
+    /// Completeness check (p.56): μ' = ν_1·μ + ⟨f, source_weights⟩ + ⟨mask_msg, mask_weights⟩
```
Bisht13: Duplicate doc comment, remove one.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,576 @@
...
+    fn check_completeness<F: Field>(r: &TestResult<F>, masked: bool) {
```
Bisht13: Tests only cover interleaving=1, num_vectors=1, Identity embedding, honest prover. Consider:
- A negative test (wrong mu or tampered proof → verifier rejects)
- A test with Basefield<Field64_2> to exercise mixed_dot/lift paths

Bisht13 (changes requested):

src/protocols/code_switch.rs
ocdbytes: we can skip this for now as I will add exhaustive testing after the completion of the protocol. As for this implementation I feel that the tests are exhaustive enough to check the completeness.

src/protocols/code_switch.rs
ocdbytes: Added a fix for this in this PR itself as it is just recomputing the params

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,593 @@
...
+    /// Queries the next stage makes to the target oracle g.
+    pub target: usize,
+    /// Queries the next stage makes to the mask oracle s.
+    pub mask: usize,
+}
+
+/// Code-switching IOR config with optional ZK.
+#[derive(Clone, PartialEq, Eq, Debug, Hash, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct Config<M: Embedding> {
```
recmo: The other files start with the `Config` struct because it is the main starting point. We should do that here too.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,593 @@
...
+    pub embedding: Typed<M>,
+    pub source: IrsConfig<M>,
+    pub target: IrsConfig<Identity<M::Target>>,
+    pub cs_ood_samples: usize,
```
recmo: The `cs_` is redundant, remove it.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,593 @@
...
+    pub mask_commit: Option<IrsConfig<Identity<M::Target>>>,
+}
+
+/// Split geometric challenge into (ν_1, rest). Requires count ≥ 1.
+fn batching_challenge<T, F>(transcript: &mut T, count: usize) -> (F, Vec<F>)
```
recmo: This is not worth the indirection, inline it and remove the function.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+    let coeffs = geometric_challenge(transcript, count);
+    let (&first, rest) = coeffs
+        .split_first()
+        .expect("count > 0 guarantees non-empty coefficients");
+    (first, rest.to_vec())
+}
+
+impl<M: Embedding> Config<M> {
+    /// `zk`: `None` for non-ZK, `Some(budget)` for ZK with next stage's query budgets.
+    pub fn new(
```
recmo, line 81: The new paper contains a new and better analysis of soundness than what is currently implemented in WHIR. This should be combined with some auxiliary constraints (choice of embedding, memory limit, security target) and costs (proof size, prover time, verifier time, etc) to create a parameter optimizer.

I consider this to be a separate next task, after we implement the protocols themselves. So while I don't mind this `new` fn being here now, I think it's a bit premature and I wouldn't spend much time on parameter selection ATM.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            mask_commit,
+        }
+    }
+
+    #[cfg_attr(feature = "tracing", instrument(skip_all))]
+    pub fn prove<H, R>(
+        &self,
+        prover_state: &mut ProverState<H, R>,
+        message: Vec<M::Target>,
+        source_randomness: &[M::Source],
```
recmo: Needs a more descriptive name. Also `source` is redundant.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            cs_ood_samples,
+            mask_commit,
+        }
+    }
+
+    #[cfg_attr(feature = "tracing", instrument(skip_all))]
+    pub fn prove<H, R>(
+        &self,
+        prover_state: &mut ProverState<H, R>,
+        message: Vec<M::Target>,
```
recmo, line 200: We currently use both `vector` and `message` for this. Within code context `vector` makes sense, so I guess for codeswitch it also does.
Once we get into inner products (sumcheck, whir) the naming `vector` and `covector` makes more sense. 

No action here, but at some point we should clean up this naming.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+        assert!(
+            self.mask_commit.is_some() == (self.target.mask_length > 0),
+            "mask config and target mask_length must agree"
+        );
+
+        // Step 1a: g := Enc_{C'}(f, r')
+        let target_witness = self.target.commit(prover_state, &[&message]);
+
+        // Step 1b: s := Enc_{C_zk}((r || padding), r'')
+        let (mask_message, mask_witness) =
```
recmo: ` = if let Some(mask_commit) = self.mask_commit {` ...

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+};
+
+/// Prover output from the code-switch.
+#[derive(Clone, Debug)]
+pub struct Witness<F: Field, G: Field = F> {
+    pub target_witness: IrsWitness<F>,
+    pub message: Vec<F>,
+    pub source_evaluations: IrsEvaluations<G>,
+    pub mask_witness: Option<IrsWitness<F>>,
+    pub mask_message: Option<Vec<F>>,
```
recmo: `pub mask: Option<(Vec<F>, IrsWitness<F>)>`

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            let f_eval = univariate_evaluate(&message, point);
+            if source_randomness.is_empty() {
+                // Non-ZK: codeword encodes only f, so y_i = f̂(ρ)
+                prover_state.prover_message(&f_eval);
+            } else {
+                // ZK: codeword encodes h(X) = f̂(X) + X^ℓ · r̂(X), so y_i = h(ρ)
+                let mask_msg = mask_message
+                    .as_ref()
+                    .expect("ZK code-switch requires mask_message");
+                let randomness_eval = univariate_evaluate(mask_msg, point);
```
recmo: `randomness` is not a descriptive name. `mask` is better, but there are so many masks here that it is still confusing. How about `mask_msg_eval` ?

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+                let shift = point.pow([msg_len as u64]);
+                prover_state.prover_message(&(f_eval + shift * randomness_eval));
+            }
+        }
+
+        // Step 4: in-domain queries
+        let source_evaluations = self.source.open(prover_state, &[source_witness]);
+
+        // Step 5: batching (advance transcript to stay in sync with verify)
+        batching_challenge::<_, M::Target>(
```
recmo: Inline this function.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            mask_message,
+        }
+    }
+
+    #[cfg_attr(feature = "tracing", instrument(skip_all))]
+    pub fn verify<H>(
+        &self,
+        verifier_state: &mut VerifierState<H>,
+        mu: M::Target,
+        source_commitment: &IrsCommitment<M::Target>,
```
recmo: `source_` is redundant.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+        let ood_points: Vec<M::Target> = verifier_state.verifier_message_vec(self.cs_ood_samples);
+        let ood_answers: Vec<M::Target> =
+            verifier_state.prover_messages_vec(self.cs_ood_samples)?;
+
+        // Step 4: source opening
+        let source_evaluations = self.source.verify(verifier_state, &[source_commitment])?;
+
+        // Step 5: batching
+        let num_ood = ood_points.len();
+        let num_in_domain = source_evaluations.matrix.len();
```
recmo, line 340: Better to get these from `self`.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+        let (original_sl_coeff, all_rlc_coeffs) =
+            batching_challenge(verifier_state, 1 + num_ood + num_in_domain);
```
recmo: Also inline here. 2 duplicate lines is much more readable than an entire new function the reader needs to learn.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+    struct TestResult<F: Field> {
+        claim: CodeSwitchClaim<F>,
+        message: Vec<F>,
+        target_mu: F,
+        mask_message: Option<Vec<F>>,
+    }
+
+    fn run_code_switch<F: Field + FieldWithSize + Codec<[u8]> + 'static>(
```
recmo: Follow the exact pattern the other protocols use: Take in only a `seed` and a `config`. The `config` can be anything that is valid, even if it is not sound or zk.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+        seed: u64,
+        size: usize,
+        src_lir: usize, // source log inverse rate
+        tgt_lir: usize, // target log inverse rate
+        masked: bool,
+    ) -> TestResult<F>
```
recmo: Test should not return a result. It should simply assert all properties the protocol guarantees.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            assert_eq!(info.ood_rlc_coeffs.len(), info.weights.ood.len());
+            assert_eq!(
+                info.in_domain_rlc_coeffs.len(),
+                info.weights.in_domain.len()
+            );
+        }
+    }
+
+    /// Powers of 2 in [8, 256] that are valid NTT domain sizes for field F.
+    fn valid_sizes<F: 'static>() -> Vec<usize> {
```
recmo: We have a helper function for this already, find it and re-use it here.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            .collect()
+    }
+
+    fn proptest_completeness<F: Field + FieldWithSize + Codec<[u8]> + 'static>()
+    where
+        Standard: Distribution<F>,
+        Hash: ProverMessage<[u8]>,
+    {
+        let rates = select(vec![1_usize, 2, 3]);
+        proptest!(|(seed: u64, sz in select(valid_sizes::<F>()), s in rates.clone(), t in rates, m: bool)| {
```
recmo: Create a `Config::arbitrary` like the other protocol tests do. Make it as adverserial as possible: any valid config can be produced (up to some reasonable size).

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,606 @@
...
+            check_completeness(&run_code_switch::<F>(seed, sz, s, t, m), m);
+        });
+    }
+
+    #[test]
+    fn completeness_field64() {
```
recmo: we can't really test for soundness or zero-knowledge. So `completeness` is redundant.

recmo (changes requested):
Most import is we need need a proptest based on arbitrary config testing all invariants.

src/protocols/code_switch.rs
ocdbytes: I think that for code switch protocol name message is more appropriate ?

src/protocols/code_switch.rs
ocdbytes: clippy flags it

src/protocols/code_switch.rs
ocdbytes: `num_in_domain` cannot be computed from self when `deduplicate_in_domain` is set as true

src/protocols/code_switch.rs
ocdbytes: I felt that in code switch providing source_* emphasises the role of the protocol and how it operates.

src/protocols/code_switch.rs
ocdbytes: wont this cause confusion about what this is ? I think it is better to make it a separate struct instead of making a tuple kind of structure. wdyt ?

src/protocols/code_switch.rs
ocdbytes: makes sense

src/protocols/code_switch.rs
ocdbytes: I felt that in code switch providing source_* emphasises the role of the protocol and how it operates. let me know wdyt

src/protocols/code_switch.rs
ocdbytes: renamed it to test_* as it seems more in tune with other protocol. But here we are just checking that the resulting claim is what it is supposed to be both in zk and non zk case. I think that counts for completeness.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,510 @@
...
+            .collect();
+        let all_points: Vec<_> = ood_points.iter().chain(&eval_points).copied().collect();
+        let mut pows: Vec<_> = ood_rlc_coeffs
+            .iter()
+            .chain(in_domain_rlc_coeffs)
+            .copied()
+            .collect();
+        for c in &mut *covector {
+            let mut sum = M::Target::ZERO;
+            for (pow, &point) in pows.iter_mut().zip(&all_points) {
```
Bisht13: This zip() truncates. pows contains one in-domain coefficient per opened matrix entry (see line 240), but all_points contains one in-domain point per opened row (lines 248–253). When source.num_messages() > 1, extra in-domain coefficients are silently dropped here, while the verifier-side μ' update still uses all of them (lines 313–325).

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,510 @@
...
+    use proptest::{bool, prelude::Strategy, proptest, sample::select, strategy::Just};
+
+    use super::*;
+    use crate::{
+        algebra::{embedding::Identity, fields, ntt, random_vector},
+        transcript::{codecs::U64, DomainSeparator},
+    };
+
+    impl<M: Embedding> Config<M> {
+        pub fn arbitrary(embedding: M) -> impl Strategy<Value = Self>
```
Bisht13, line 388: These proptests build Config directly instead of going through Config::new(), so they do not validate the constructor-derived invariants. Since this PR adds nontrivial constructor logic, the main property tests should exercise Config::new() rather than bypass it.

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,510 @@
...
+        );
+        let proof = prover_state.proof();
+
+        let mut verifier_state = VerifierState::new_std(&ds, &proof);
+        let source_commitment = config
+            .source
+            .receive_commitment(&mut verifier_state)
+            .unwrap();
+        let mut verifier_sum = target_mu;
+        let commitments = config
```
Bisht13, line 482: The PR description says the tests include a completeness check for the paper’s μ' decomposition, but this test currently only checks transcript sync and mask presence/absence after verify(). I don’t see a direct assertion here that the prover-side covector update and verifier-side μ' update match the Construction 9.7 completeness identity.

Bisht13 (changes requested):

src/protocols/code_switch.rs
ocdbytes: Right now config for code switch is not solidified as there are parameters which are not accounted for right now (This will be done in the future PR which specifically will handle the parameter selection).

src/protocols/code_switch.rs
ocdbytes: Updated description for the tests that will be added in future PRs when orchestrator is implemented. As right now we can't test the proper completeness without making testing logic complicated.

src/protocols/code_switch.rs
ocdbytes: Code switch only supports one vector which means that this case will never arrive. To solidify the logic I have added asserts. for both prove and verify

Bisht13 (approved):

src/protocols/irs_commit.rs
```diff
@@ -332,6 +345,8 @@ impl<M: Embedding> Config<M> {
         let matrix_witness = self.matrix_commit.commit(prover_state, &matrix);
 
         // Handle out-of-domain points and values
+        // TODO : Remove this logic after main whir protocol is updated
```
WizardOfMenlo, line 348: Not sure we can actually remove this TODO unfortunately

src/protocols/irs_commit.rs
```diff
@@ -245,6 +237,27 @@ impl<M: Embedding> Config<M> {
         self.out_domain_samples == 0 && self.johnson_slack == 0.0
     }
 
+    /// Recompute `johnson_slack`, `in_domain_samples`, and `out_domain_samples`
+    /// for the current rate (which accounts for `mask_length`).
+    pub fn reparameterise_security(&mut self, security_target: f64, unique_decoding: bool) {
```
WizardOfMenlo: Can we rename this to something more informative? This basically computes the parameters for this step right?

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,523 @@
...
+                security_target,
+                unique_decoding,
+                hash_id,
+                1,
+                source_randomness_len,
+                1,
+                source_config.rate(),
+            );
+            mask_config.mask_length = budget.mask;
+            mask_config.reparameterise_security(security_target, unique_decoding);
```
WizardOfMenlo: Can we make reparametrise security be done in the new of mask_config? This makes it harder to misuse

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,523 @@
...
+            self.mask_commit.is_some() == (self.target.mask_length > 0),
+            "mask config and target mask_length must agree"
+        );
+
+        // Step 1a: g := Enc_{C'}(f, r') — Construction 9.7 Step 1, p.55
+        let target_witness = self.target.commit(prover_state, &[&message]);
+
+        // Step 1b: s := Enc_{C_zk}((r || padding), r'') — Construction 9.7 Step 1, p.55
+        #[allow(clippy::option_if_let_else)]
+        let mask = if let Some(mask_config) = &self.mask_commit {
```
WizardOfMenlo, line 234: Can we make this a map instead? (Instead of if else)

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,523 @@
...
+            let witness = mask_config.commit(prover_state, &[&mask_msg]);
+            Some(MaskOracle {
+                message: mask_msg,
+                witness,
+            })
+        } else {
+            None
+        };
+
+        // Step 2-3: OOD challenge + answers — Construction 9.7 Steps 2-3, p.55
```
WizardOfMenlo, line 252: We should double check that the current univariate_eval satisfies the private zero-evader conditions that we have in the paper, I am not sure if they do

WizardOfMenlo: Let's leave a comment about it

src/protocols/code_switch.rs
```diff
@@ -0,0 +1,523 @@
...
+                prover_state.prover_message(&(f_eval + shift * mask_msg_eval));
+            } else {
+                prover_state.prover_message(&f_eval);
+            }
+        }
+
+        // Step 4: in-domain queries — Construction 9.7 Step 4, p.55
+        let source_evaluations = self.source.open(prover_state, &[witness]);
+
+        // Step 5: batching — Construction 9.7 Step 5, p.55
```
WizardOfMenlo: This is still step 4 right?

WizardOfMenlo (approved):
Approved, left some comments. The most annoying footgun is that we should check what is the right zero-evader to use. 

Here is what AI had to say about it:


>  ### [P2] src/protocols/code_switch.rs:207 — target_interleaving_depth > 1 makes    proving panic
>                                                                                     Config::new allows any positive target_interleaving_depth and sets the target IRS
>  vector_size to source_message_length * target_interleaving_depth, but prove still  passes a vector whose length is only source_message_length into
>  self.target.commit. For any caller using the advertised target_interleaving_depth  > 1, IrsConfig::commit rejects the input length and panics. Either restrict this
>  config to depth 1 or pass a target vector with the length/layout expected by the   target IRS config.
>                                                                                     ### [P2] src/protocols/code_switch.rs:118 — ZK mask lengths can create invalid
>  IRS configs
>  In ZK mode, target_config.mask_length and mask_config.mask_length are assigned
>  after IrsConfig::new has already chosen codeword_length from the unmasked message  length. If budget.target or budget.mask exceeds the spare capacity implied by the
>  original rate, masked_message_length() > codeword_length, so later commit/open     calls panic in the NTT evaluation path; even before that, reparameterise_security
>   computes rates above 1. This arises for valid-looking inputs such as rate 1/2     with a mask budget larger than the message length. The constructor should either
>  size codeword_length for message_length + mask_length or reject budgets that do    not fit.
>  ### [P2] src/protocols/code_switch.rs:267 — ZK covector update omits the
>  mask-oracle coefficients
> 
>  In ZK mode, the verifier updates sum using OOD answers and source openings for
>  the combined polynomial f(x) + x^n s(x), but the prover-side covector update only
>  mutates the caller-provided single covector and does not require or return the
>  mask portion corresponding to s. With the f-length covector used by the current
>  test and implied by message.len(), honest proofs with nonzero source masks will
>  produce a verifier sum that includes mask terms while the downstream target
>  relation only has weights for f. Require/update a covector of length
>  message.len() + mask_message_len and expose the split for target/mask, or keep
>  the verifier update f-only.
> 
>  Overall Verdict
> 
>  needs attention

src/protocols/irs_commit.rs
ocdbytes: Why ? we can remove this right we dont need ood points in commitment after we update the protocol as this should be handled by code switch. Here it is just revealing more info about the vector.

src/protocols/code_switch.rs
ocdbytes: Yeah so the thing is this work will be cleaned up in parameter selection PR. here I just added for the correctness of the protocol. The configs are very messed up in each of the sub protocols for eg : 
in IRS commit the way mask length is being handled is bad design. Will fix this I have noted these things on my end.

src/protocols/code_switch.rs
ocdbytes: It is a design choice, initially it was map and remco pointed out to use the if let pattern. Ig this looks clean even when we need to suppress clippy suggestion of map. with map it looked like : 

```rust
        let (mask_message, mask_witness) =
            self.mask_commit
                .as_ref()
                .map_or((None, None), |mask_config| {
                    let mask_msg_len = mask_config.message_length();
                    let r_embedded = lift(self.source.embedding(), source_randomness);
                    let embedded_randomness_len = r_embedded.len();
                    let mut mask_msg = Vec::with_capacity(mask_msg_len);
                    mask_msg.extend_from_slice(&r_embedded);
                    let random_padding: Vec<M::Target> =
                        random_vector(prover_state.rng(), mask_msg_len - embedded_randomness_len);
                    mask_msg.extend_from_slice(&random_padding);
                    let witness = mask_config.commit(prover_state, &[&mask_msg]);
                    (Some(mask_msg), Some(witness))
                });
```

src/protocols/code_switch.rs
ocdbytes: makes sense

src/protocols/code_switch.rs
ocdbytes: yup according to the paper yes. I just divided into different step for batching part.

ocdbytes:
@WizardOfMenlo 

Agreed on the deterministic evader — paper has a private ze_ood(ρ) but the geometric progression (1, α, α², …) works fine: applied to [f; (r,s)] it just evaluates the concatenated polynomial at α. ZK still holds because y = f(α) + α^ℓ · (r,s)(α) is uniform whenever (r,s) has fresh randomness and α ≠ 0. The α = 0 event is 1/|F| and folds into the soundness budget. So as long as ℓ_zk - r ≥ t_ood we get perfect privacy ζ_ze = 0.

Soundness parameter derivation pending.

src/protocols/irs_commit.rs
ocdbytes: how about : recompute_security_parameters ?

src/protocols/irs_commit.rs
ocdbytes: related slack message link : https://atheonlabs.slack.com/archives/C083C5JM8QH/p1775500051740209

PR #251, opened 2026-04-29
zkWHIR 3.0 : Updated protocols for whir round integration and added mask proximity protocol (part 2)

ocdbytes:
## Summary

- Add mask proximity protocol (Construction 7.2) — verifies mask polynomials are close to C_zk codewords via shared Merkle tree
- Refactor code-switch (Construction 9.7) — ZK-correct OOD with [f; r; s], covector split, mask oracle externalized to orchestrator
- Update sumcheck (Construction 6.3) — unified output type, ZK masking support
- Update basecase — structured Opening<F> output type
- Prepare all sub-protocols for orchestrator integration with shared mask tree per round

## Mask Proximity Protocol

Implements Construction 7.2 from the paper (§7, p.43-48), specialized for zero-constraint masks (μ_i = 0, sl_{o,i} = 0).

#### Protocol flow:
  1. Prover commits 2n vectors in one tree: n original masks ξ_i and n fresh mask-of-masks s_i
  2. Verifier sends γ (combination randomness)
  3. Prover sends combined polynomials ξ*_i = s_i + γ·ξ_i and combined IRS randomness r*_i = r'_i + γ·r_i
  4. Shared tree is opened at random positions
  5. Verifier spot-checks: Enc(ξ*_i, r*_i)(y_j) = s_i(y_j) + γ·ξ_i(y_j) using linearity of RS encoding

> [!IMPORTANT]
> Soundness (Lemma 7.4, p.45): If ξ_i is δ-far from every C_zk codeword, then ξ*_i = s_i + γ·ξ_i is also far from s_i + γ·c for every codeword c, with high probability over γ. The spot-check catches disagreement at random positions with probability ≥ 1 − (1−δ)^{t_zk}.

> [!IMPORTANT]
> ZK safety (Lemma 7.3, p.44): Only ξ*_i = s_i + γ·ξ_i is revealed in full. Since s_i is uniformly random, ξ*_i is uniform regardless of ξ_i — the original mask is never exposed. The tree is opened at ≤ t_zk positions, within the ZK query budget of C_zk's encoding.

> [MISSED IN DISCUSSIONS]
> Why not just irs_commit.open/verify: irs_commit.verify() only checks Merkle binding (authentication paths). It does NOT verify proximity to C_zk — the doc comment in irs_commit confirms: "This does not verify the out-of-domain evaluations." Proximity requires the γ-combination: send the combined polynomial in full, then spot-check against the committed tree.

## Next Steps

- Protocol Param Selection
- Wiring of protocol


src/protocols/code_switch.rs
```diff
@@ -15,192 +15,139 @@ use crate::{
...
+        assert_eq!(
+            target_config.vector_size,
+            source_config.message_length(),
+            "target vector_size must equal source message_length (target encodes one polynomial of length ℓ)"
+        );
+        assert!(
+            target_config.interleaving_depth.is_power_of_two(),
+            "target.interleaving_depth must be a power of 2"
+        );
+        // Theorem 9.6: ℓ_zk ≥ r (mask oracle must cover source randomness).
```
Bisht13, line 93: This validates only `ℓ_zk ≥ r`. For the current OOD map `f(α) + α^ℓ·mask(α)`, Lemma 9.8 also needs a checked privacy/rank condition for `ze_ood`; otherwise configs with too little fresh `s` padding can be accepted. Please reject under-padded ZK configs here, or make this constructor explicitly unchecked until parameter selection enforces the condition.

src/protocols/code_switch.rs
```diff
@@ -269,36 +207,57 @@ impl<M: Embedding> Config<M> {
...
 
-    /// Verify the code-switch
+    /// Verify the code-switch.
+    ///
+    /// `folding_randomness` is the **sumcheck folding randomness `γ`** the
+    /// verifier derived from the transcript during the preceding sumcheck.
+    /// It must match what the prover received from the same transcript —
+    /// not caller-supplied randomness. See `prove` doc for details.
+    ///
+    /// Returns the target commitment. In ZK mode, the caller **must**
```
Bisht13, line 261: This is soundness-critical and should not be a doc-only caller obligation. Construction 9.7’s next relation includes the mask oracle `s`; `code_switch::verify` currently returns only `g`. Please make ZK code-switch return/accept a typed next-claim including the mask commitment, or provide a composed verifier that runs code-switch + mask-proximity in a fixed transcript order.

src/protocols/code_switch.rs
```diff
@@ -213,52 +160,43 @@ impl<M: Embedding> Config<M> {
...
+                    mask.len(),
+                    self.message_mask_length,
+                    "mask_msg length must equal message_mask_length"
+                );
+                Some(mask)
+            }
+        };
 
-        // Step 1a: g := Enc_{C'}(f, r') — Construction 9.7 Step 1, p.55
+        // Step 1: g := Enc_{C'}(f, r') — Construction 9.7 Step 1, p.55
```
Bisht13, line 197: In ZK mode, `target.commit` must itself be a ZK-safe encoding of `g`. The constructor currently allows `target.mask_length == 0` and/or `target.out_domain_samples > 0`, which does not satisfy Lemma 9.8’s `C′` ZK-encoding assumption and may leak raw target evaluations. Please validate the target config in ZK mode or make this an explicitly unchecked low-level primitive.

src/protocols/mask_proximity.rs
```diff
@@ -0,0 +1,515 @@
...
+            .c_zk_commit
+            .open(&mut prover_state, &[&witness.mask_witness]);
+        let proof = prover_state.proof();
+
+        let mut verifier_state = VerifierState::new_std(&ds, &proof);
+        let commitment = config.receive_commitment(&mut verifier_state).unwrap();
+        assert_rejected(|| config.verify(&mut verifier_state, &commitment));
+    }
+
+    fn assert_rejected<F>(verify: impl FnOnce() -> VerificationResult<F>) {
```
Bisht13: This test accepts panic as a valid rejection. In normal builds malformed proofs should return `VerificationError`; panic should only be accepted under `verifier_panics`. Otherwise verifier error handling can regress while the tamper test still passes.

src/protocols/sumcheck.rs
```diff
@@ -66,47 +74,45 @@ impl<F: Field> Config<F> {
...
         masks: &[F],
-    ) -> (Vec<F>, F, F)
+    ) -> SumcheckOpening<F>
     where
         H: DuplexSpongeInterface,
         R: CryptoRng + RngCore,
         F: Codec<[H::U]>,
         [u8; 32]: Decoding<[H::U]>,
         U64: Codec<[H::U]>,
+        Standard: Distribution<F>,
```
Bisht13: These new bounds appear unused: `Standard: Distribution<F>` and `Hash: ProverMessage<[H::U]>` on `prove`, plus `Hash: ProverMessage<[H::U]>` on `verify`. I checked that the library still compiles without them. Could we drop the extra bounds to keep the API surface minimal?

Bisht13 (changes requested):

src/protocols/code_switch.rs
ocdbytes: Added the assert here but would remove it in the parameter selection PR

src/protocols/code_switch.rs
ocdbytes: The idea is to have orchestrator own the masks and commit them. pass the refs in sub protocols and if there is tampering the mask proximity check should fail.

src/protocols/mask_proximity.rs
```diff
@@ -0,0 +1,525 @@
...
+        assert_eq!(original_msgs.len(), self.num_masks);
+        assert_eq!(witness.fresh_msgs.len(), self.num_masks);
+
+        // Step 1: receive combination randomness γ
+        let gamma: F = prover_state.verifier_message();
+
+        // Step 2: compute and send combined polynomials + IRS randomness
+        let irs_masks_per_vector =
+            self.c_zk_commit.mask_length * self.c_zk_commit.interleaving_depth;
+        debug_assert_eq!(
```
Bisht13: Use assert_eq!

src/protocols/code_switch.rs
```diff
@@ -309,65 +292,60 @@ impl<M: Embedding> Config<M> {
         Hash: ProverMessage<[H::U]>,
     {
         assert_eq!(
```
Bisht13: verify uses assert_eq! for folding_randomness shape. Library verifiers should return VerificationError, not panic. Convert to verify! or structured verifier error.

src/protocols/basecase.rs
```diff
@@ -63,7 +70,10 @@ impl<F: Field> Config<F> {
...
         debug_assert_eq!(dot(&vector, &covector), sum);
         if self.size() == 0 {
-            return (Vec::new(), F::ZERO);
+            return Opening {
+                evaluation_points: Vec::new(),
+                linear_form_evaluation: F::ZERO,
+            };
         }
 
         // Even more trivial non-zk protocol: send f an r directly.
```
Bisht13: Nit:  “send f an r directly” → “send f and r directly”

src/protocols/mask_proximity.rs
```diff
@@ -0,0 +1,525 @@
...
+#[derive(Clone, PartialEq, Eq, Debug, Hash, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct Config<F: Field> {
+    pub c_zk_commit: IrsConfig<Identity<F>>,
+    pub num_masks: usize,
+}
+
+/// Prover output from the commit phase.
+#[must_use]
+pub struct Witness<F: Field> {
```
Bisht13, line 73: Witness::mask_witness and Witness::fresh_msgs are public. External mutation between commit and prove can cause invalid proofs. Not a soundness leak, but avoidable API footgun. Make fields private or pub(crate).

src/protocols/basecase.rs
```diff
@@ -74,9 +84,12 @@ impl<F: Field> Config<F> {
             let point = self
                 .sumcheck
                 .prove(prover_state, &mut vector, &mut covector, &mut sum, &[])
-                .0;
+                .round_challenges;
             assert!(!vector[0].is_zero(), "Proof failed");
```
Bisht13, line 88: Here and at :107, :135. These three `assert!(_, "Proof failed")` panic the prover on (a) all-zero input vector, (b) `mask_rlc == 0` (1-in-2^64), (c) zero post-fold masked vector. An honest prover hitting any of these in deployment should `Err` and let upstream re-randomize, not crash. Consider a `pub enum ProverError { Degenerate, ChallengeIsZero }` and return `Result<Opening<F>, ProverError>`.

src/protocols/sumcheck.rs
```diff
@@ -77,36 +84,32 @@ impl<F: Field> Config<F> {
...
         assert_eq!(a.len(), self.initial_size);
         assert_eq!(b.len(), self.initial_size);
         debug_assert_eq!(dot(a, b), *sum);
         assert_eq!(masks.len(), self.num_rounds * self.mask_length);
         let half = F::from(2).inverse().unwrap();
 
-        // Send mask sum and get combination randomness.
         let mut mask_sum = F::ZERO;
         let mut mask_rlc = F::ONE;
         if !masks.is_empty() {
```
Bisht13: Here and at :183. Prover gates ZK path on `!masks.is_empty()` here; verifier gates on `mask_length > 0 && num_rounds > 0` at line 183. Equivalent only because of the `assert_eq!(masks.len(), num_rounds * mask_length)` at line 92. Can we pick one expression and use it on both sides, preferably `mask_length > 0 && num_rounds > 0` since the verifier doesn't have `masks` in scope. Reduces the chance of a future refactor desynchronizing them silently.

src/protocols/code_switch.rs
```diff
@@ -15,192 +15,148 @@ use crate::{
...
             source: source_config,
             target: target_config,
+            message_mask_length,
             out_domain_samples,
-            mask_commit,
         }
     }
 
     /// Length of the covector for this code-switch.
     pub fn covector_length(&self) -> usize {
```
Bisht13, line 128: nit: the `.max()` is dead code given the asserts at lines 95 (`message_mask_length >= source.mask_length` when ZK) and 110 (`source.mask_length == 0` when not ZK). Either of these forces `message_mask_length >= source.mask_length` always, so this can be `+ self.message_mask_length`.

Bisht13 (approved):

src/protocols/basecase.rs
WizardOfMenlo: Wait why? The all-zero input vector is a valid input vector.

WizardOfMenlo (approved):

PR #257, opened 2026-06-15
Zook : Parameter Selection 

ocdbytes:
**Changes :** 
- Added proper param selection.
     - All paths for param selection are protected using type safety.
     - Any individual config cannot be modified within param selection.
     - Added analytical bit helper for each sub protocol param selection which also accounts for security bits.
     - Derived bounds for our modfied params can be found here : [Derived Bounds](https://hackmd.io/@1q1q-TiuQN6fAkxaN41u-Q/ryBoT_UA-e).
     - All the bounds are verified by @WizardOfMenlo 
- Added Mode enum for each sub protocol for cleaner branching of the execution flow. 
- POW Grinding added for all the sub protocols.
- Added getters instead of pub access of config fields for safety.

**Tests :** 
- Added proptests for each param selection path and protocol.

This PR was squashed from 47 commits during conflict resolution against PR #251  (merged to `main`)
Commit history maintained in personal fork.


shreyas-londhe:
@ocdbytes Can you please resolve the merge conflicts on this PR? I'm reviewing it and we are planning to merge it by this week.

zkfriendly:
Hey @shreyas-londhe, this branch did not have any conflicts with main before merging buffer abstraction. I opened a PR to this branch here https://github.com/ocdbytes/whir/pull/11, which adds buffer abstraction. Merging it will make this branch conflict free again. cc @ocdbytes

ocdbytes:
Thanks @zkfriendly I have merged your PR now there are no conflicts. I skimmed through the changes will go through them properly tomorrow to see if there is anything to be changed. 
cc @shreyas-londhe

src/protocols/params/protocol_config.rs
```diff
@@ -0,0 +1,522 @@
...
+    },
+};
+
+#[derive(Clone, Debug)]
+pub struct ProtocolConfig<M: Embedding> {
+    security: SecuritySpec,
+    tuning: TuningSpec,
+    rounds: Vec<RoundConfig<M>>,
+    basecase: BasecasePlan<M::Target>,
+}
```
BornPsych, line 37: Can `ProtocolConfig` derive `Serialize`/`Deserialize`? It's `#[derive(Clone, Debug)]` today, so nothing downstream can embed it in a serialized scheme.

Context for why I'm asking: ProveKit builds its Fiat-Shamir domain separator by serializing the whole scheme (WHIR config included) and hashing that. `whir::Config` already handles it, with `Serialize`/`Deserialize` and `#[serde(bound = "")]` for the embedding generic. Once this stack has a prover/verifier runtime and replaces `whir::Config` as what schemes carry, `ProtocolConfig` will need the same.

It's not a one-liner though. `RoundConfig`, `RoundMode`, `MaskOracleConfig`, `BasecasePlan` in this file, plus `Solved` and `CodeSwitchConfig`, all need it too, and none of them derive it yet (`SecuritySpec` and `TuningSpec` already do). The embedding-generic ones want `#[serde(bound = "")]` like `whir::Config`.

ocdbytes: Hey @BornPsych 
I have added these derives in wiring PR (next PR). Ig missed it here in this PR

src/protocols/params/regime.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn eps_mca_log2_johnson_formula() {
+        let canonical_slack = 2_f64.powf(-MCA_LOG_INV_RATE).sqrt() / 20.0;
+
+        let got = johnson(canonical_slack).eps_mca_log2(
+            MCA_LOG_INV_RATE,
+            MCA_MESSAGE_LENGTH,
+            MCA_FIELD_BITS,
+        );
+        let bcss25_const = (2.0 * 10.5_f64.powi(5) / 3.0).log2();
+        let expected = bcss25_const + (MCA_MESSAGE_LENGTH as f64).log2() + 2.5 * MCA_LOG_INV_RATE
```
shreyas-londhe: **issue (non-blocking):** `expected` is built from the same expression as the implementation, so this test can't catch a transcription error in the bound itself.

A wrong constant or exponent in `eps_mca_log2` would update both sides equally and the test would stay green. Since this file is where the soundness math lives, I'd anchor at least one case per regime to a hand-computed number: fix k, ρ, and |F|, work out the bits from the theorem by hand, and assert against that literal with the derivation in a comment. The `analytic_error_*_formula` tests in the solver modules and `privacy_error_bits_matches_bound_3_sum` in `derive.rs` have the same shape.

src/protocols/irs_commit.rs
```diff
@@ -577,51 +607,42 @@ impl<M: Embedding> fmt::Display for Config<M> {
...
-    } else {
-        // Johnson bound: δ = 1 - √ρ - η
-        rate.sqrt() + johnson_slack
-    };
-    (security_target / (-per_sample.log2())).ceil() as usize
+) -> NonZeroUsize {
+    let regime = DecodingRegimeParams::from_policy(decoding_regime, rate);
+    // Query error is (1 - δ)^q in bits = -q · log2(1 - δ).
+    let log_one_minus_delta = regime.one_minus_distance_log2(-rate.log2());
+    let q = (security_target / -log_one_minus_delta).ceil() as usize;
```
shreyas-londhe, line 655: **issue (non-blocking):** `num_in_domain_queries` divides by zero when `rate == 1`.

At `log_inv_rate = 0` the Unique branch gives `1 − δ = 1`, so `log_one_minus_delta` is 0 and `q` saturates to `usize::MAX` through the `ceil() as usize` cast. `TuningSpec` doesn't reject `starting_log_inv_rate == 0` and `-r 0` sets it directly from the CLI, so this is reachable from the binary. A `starting_log_inv_rate ≥ 1` check in `round_layout` (or a new `LayoutError` variant) would turn this into a clean error.

src/protocols/params/regime.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// - Unique: `(k − 1) / |F|`, log = `log k − |F|` (with `+ log ρ⁻¹`).
+    /// - Johnson: BCSS25 Theorem 1.5 at canonical `η = √ρ/20`, `m = 10`:
+    ///   `ε ≈ (2·10.5⁵/3) · n · ρ^{−3/2} / |F|`.
+    /// - Capacity: STIR Conj 5.6, `ε ≈ d / (η · ρ²) / |F|`.
+    pub fn eps_mca_log2(self, log_inv_rate: f64, message_length: usize, field_bits: f64) -> f64 {
+        let log_k = usize_to_f64(message_length).log2();
+        let error = match self {
+            Self::Unique => log_k + log_inv_rate,
+            Self::Johnson { slack } => {
+                debug_assert!(
```
shreyas-londhe: **suggestion (non-blocking):** keep these as hard `assert!`s like the pre-PR code.

The baseline enforced the η lower bound with `assert!`, so release builds kept the backstop. The `DecodingRegimeParams` variants are `pub`, so a future caller can construct `Johnson { slack }` directly with an out-of-range η and silently get an over-optimistic error bound in release. The check is two float ops on a cold path, so keeping it hard costs nothing. Same applies to the Capacity guard on line 121.

shreyas-londhe (approved):

src/protocols/params/regime.rs
ocdbytes: added theorem tests for more concrete testing from actual formulas from literature

src/protocols/irs_commit.rs
ocdbytes: I don't exactly get how `q` saturates to usize::MAX but there was some other issue which I saw while checking this so added a proper guard here. If `q == 1` then it produces params with no soundness although it is not possible when the guard for starting rate should be `>=1`. So it is a soundness issue mitigated in appropriate places.

src/protocols/params/regime.rs
ocdbytes: added proper guard here

PR #260, opened 2026-06-19
Buffer Abstraction

zkfriendly:
## Summary

This PR introduces the buffer abstraction layer. All buffer abstraction traits are defined in the [buffer crate](https://github.com/zkfriendly/whir/blob/d68e838227176eb596f47799d0543ffdee75d579/src/buffer/mod.rs#L45), with a CPU implementation that essentially wraps the existing in-memory code paths.

At this moment, the buffer crate still depends on existing algebra and protocol crates. Future work can consolidate the relevant logic directly into the buffer crate as the abstraction settles.

## Notes

There is also an experimental Metal implementation, but it is not included in this PR.

`zk_whir` is not fully ported to the buffer abstraction yet. It uses `.as_slice()` in places to keep the code working, so behavior is essentially unchanged compared to directly using CPU buffers. Since WHIR v3 is coming and is expected to replace `zk_whir`, that path is intentionally not fully ported for now.


zkfriendly:
cc @recmo

src/protocols/basecase.rs
```diff
@@ -78,38 +79,47 @@ impl<F: Field> Config<F> {
...
             let _ = self.commit.open(prover_state, &[witness]);
             let point = self
                 .sumcheck
                 .prove(prover_state, &mut vector, &mut covector, &mut sum, &[])
                 .round_challenges;
-            assert!(!vector[0].is_zero(), "Proof failed");
+            assert!(
+                !vector.at_index(0).unwrap_or_else(F::one).is_zero(),
+                "Proof failed"
+            );
```
shreyas-londhe, line 93: issue (blocking): `unwrap_or_else(F::one)` makes this pass when the buffer is empty.

`F::one()` is non-zero, so an out-of-bounds `at_index(0)` falls back to a value that satisfies the check instead of failing it. The covector lines just below (96, 152) already use `.expect("Proof failed")`. Use that here too. Same fix at line 146.

src/protocols/merkle_tree.rs
```diff
@@ -78,13 +78,10 @@ impl Config {
...
-    pub fn commit<H, R>(&self, prover_state: &mut ProverState<H, R>, leaves: Vec<Hash>) -> Witness
-    where
-        H: DuplexSpongeInterface,
-        R: RngCore + CryptoRng,
-        Hash: ProverMessage<[H::U]>,
-    {
+    /// Build the full node array from leaf hashes, without touching the transcript.
+    ///
+    /// The returned vector holds the leaf layer first and the root last (at `num_nodes() - 1`).
+    pub fn build_nodes(&self, leaves: Vec<Hash>) -> Vec<Hash> {
```
shreyas-londhe, line 84: **suggestion (non-blocking):** `build_nodes` leaves the root-send as an unenforced caller duty.

Splitting construction from the transcript write is good, but sending the root is now manual (`prover_message(&nodes[num_nodes()-1])`). A future caller that forgets it produces a commitment whose root never binds, with no compile error and probably no failing test. Consider funneling `Witness` creation through a method that takes `prover_state` and sends the root, or returning a `#[must_use]` wrapper that only a transcript-taking method unwraps. At minimum, document the obligation here.

src/algebra/ntt/mod.rs
```diff
@@ -88,15 +120,20 @@ pub trait ReedSolomon<F>: Debug + Send + Sync {
...
     /// Each output value is the univariate polynomial evaluation in the evaluation point
     /// corresponding with the index of a coefficient list formed by concatenating message and mask.
-    ///
-    fn interleaved_encode(&self, messages: &[&[F]], masks: &[F], codeword_length: usize) -> Vec<F>;
+    fn interleaved_encode(
+        &self,
+        messages: Messages<'_, F>,
+        masks: &ActiveBuffer<F>,
+        codeword_length: usize,
+    ) -> ActiveBuffer<F>;
```
shreyas-londhe, line 136: **suggestion:** `ReedSolomon` is a backend-abstraction trait but hardcodes the global `ActiveBuffer` alias in its signature, which defeats the purpose for the one trait that most needs it.

Give it an associated `type Buffer: Buffer<F>` and thread it through the registry (`type Dyn<F> = dyn ReedSolomon<F, Buffer = ActiveBuffer<F>>`); `NttEngine` then declares `type Buffer = CpuBuffer<F>`. The payoff is enforcement, not tidiness: today a second backend can't even be expressed here, since the trait forces every impl to speak `CpuBuffer`. With the associated type, an impl whose buffer doesn't match the build's `ActiveBuffer` fails to register as a *type error* instead of sitting as dead code that forces host round-trips if called.

src/buffer/cpu/buffer.rs
```diff
@@ -0,0 +1,312 @@
...
+
+    fn fold_pair(&mut self, other: &mut Self, weight: F) {
+        self.fold(weight);
+        other.fold(weight);
+    }
+
+    fn fold_pair_sumcheck_polynomial(&mut self, other: &mut Self, weight: F) -> (F, F) {
+        self.fold_pair(other, weight);
+        self.sumcheck_polynomial(other)
+    }
```
shreyas-londhe: **nitpick (non-blocking):** these two overrides duplicate the trait defaults.

`fold_pair` and `fold_pair_sumcheck_polynomial` here have the same bodies as the defaults on `Buffer`.  Drop both from the impl and inherit the defaults, unless an override is planned for a reason.

shreyas-londhe (changes requested):
Thanks for the PR @zkfriendly. Really like where this lands. Trait split is the right cut, the CPU backend stays a thin wrapper with no second implementation to drift, and the prover got simpler dropping the `Any`/`Cow` recycling for `mixed_linear_combination` and `linear_forms_rlc`. Left a few inline comments, but the shape is solid.

zkfriendly:
Superseded by #263 — same branch now pushed directly to worldfnd/whir (GitHub can't change a PR's head repo, so it had to be reopened). Review comments here have been addressed on the branch: the blocking basecase check is fixed, build_nodes caller obligation documented, and the fold_pair overrides removed as part of a larger simplification that drops the slice/view layer entirely. Replies on the individual threads to follow on #263.

src/algebra/ntt/mod.rs
zkfriendly: I agree with this point. But I think we can keep it simple for now, and refactor it once we introduce another backend type.

PR #263, opened 2026-07-02
Buffer Abstraction

zkfriendly:
## Summary

This PR introduces the buffer abstraction layer. All buffer abstraction traits are defined in the buffer crate (`src/buffer/mod.rs`), with a CPU implementation that essentially wraps the existing in-memory code paths.

At this moment, the buffer crate still depends on existing algebra and protocol crates. Future work can consolidate the relevant logic directly into the buffer crate as the abstraction settles.

Since the original review, the abstraction has been simplified substantially: the slice/view layer (`BufferRead`/`BufferWrite`, `CpuSlice`/`CpuSliceMut`) is removed in favor of two traits (`BufferOps`, `Buffer`) over owned buffers only. Host readbacks (`as_slice`) now occur only at transcript boundaries where the data must reach the host anyway.

## Notes

There is also an experimental Metal implementation, but it is not included in this PR.

`zk_whir` is not fully ported to the buffer abstraction yet. It uses `.as_slice()` in places to keep the code working, so behavior is essentially unchanged compared to directly using CPU buffers. Since WHIR v3 is coming and is expected to replace `zk_whir`, that path is intentionally not fully ported for now.

---

Supersedes #260 (same branch, previously from my fork — GitHub cannot retarget a PR head, so reopened from the in-repo branch). Review history and comment threads are on #260.


src/protocols/matrix_commit.rs
shreyas-londhe: **issue (blocking):** `pub` was added only to the `parallel` variant; this one stays private while `buffer/cpu.rs:17` imports it unconditionally. `cargo check --no-default-features --lib` fails. Add `pub` here too.

src/protocols/sumcheck.rs
```diff
@@ -110,9 +107,9 @@ impl<F: Field> Config<F> {
         {
             // Fold and compute sumcheck polynomial in one pass.
             let (c0, c2) = if let Some(w) = prev_round_challenge {
-                fold_and_compute_polynomial(a, b, w)
+                a.fold_pair_sumcheck_polynomial(b, w)
             } else {
-                compute_sumcheck_polynomial(a, b)
+                a.sumcheck_polynomial(b)
```
shreyas-londhe, line 112: **suggestion:** `fold_and_compute_polynomial` in `algebra/sumcheck.rs` lost its only caller here and is now dead code. Also this "in one pass" comment is stale: the CPU path still runs three passes (fold a, fold b, compute), same as before.

src/protocols/merkle_tree.rs
```diff
@@ -58,7 +58,7 @@ pub struct Commitment {
 #[must_use]
 pub struct Witness {
     /// The nodes in the Merkle tree, starting with the leaf hash layer.
-    nodes: Vec<Hash>,
+    pub nodes: Vec<Hash>,
```
shreyas-londhe: **suggestion:** `merkle_tree::open` and `Witness` have no production caller anymore (matrix_commit now has its own `Witness` and inlines the open via `opening_sibling_indices`); only merkle_tree's own tests use them. Either drop this path or route matrix_commit back through it. Also `nodes` doesn't need to be `pub` — its only constructor is the test module, which already has private access. Same for `matrix_commit::Witness::nodes`.

shreyas-londhe (commented):
Left a few minor comments, but overall looks clean. I'm not completely happy with including Merkle Tree ops in BufferOps, but it is not a blocker and we can move it out in a later PR when the need arises. Thanks for the PR @zkfriendly

src/protocols/sumcheck.rs
zkfriendly: > fold_and_compute_polynomial in algebra/sumcheck.rs lost its only caller here and is now dead code. 

Good catch, actually we should just use `fold_and_compute_polynomial` instead of the three passes and keep the comment

zkfriendly:
> I'm not completely happy with including Merkle Tree ops in BufferOps

What about it doesn't fit into BufferOps?

shreyas-londhe:
> What about it doesn't fit into BufferOps?

It is about concern separation, we should follow a hierarchal trait system, where we compose traits instead of adding all the functionality in a single trait. Currently I have to implement the merkle tree ops even if I don't use it in code for example. But it is a non-blocker for now.

shreyas-londhe (approved):

xrvdg (approved):
Discussed offline:
- [X] Message parameter for RS
- [X] Leverage type system for mutability and slices rather than specialised types.
- [X] Naming guidelines for ad-hoc conversion

Non-blocking / follow-up PR:
- Merklize feels out of place inside BufferOps.
- We could get better ergonomics by replacing from_vec/from_slice with the standard From trait for a specific backend.
- Traits are usually best when they are composable rather than hierarchical. Since BufferOps seems to handle host <-> backend communication, and Buffer should ideally only care about backend-specific operations, we might be able to remove BufferOps as a trait dependency on Buffer.

I think after this, the follow up PRs and a bit of usage a more intuitive name for the traits will come to mind.

PR #266, opened 2026-07-13
Make buffer capabilities composable

zkfriendly:
## Summary

Refactors the buffer abstraction so backend capabilities are more composable and more protocol work can stay buffer-native.

- Replace custom `from_vec` / `from_slice` constructors with standard `From<Vec<T>>` and `From<&[T]>` implementations.
- Split host-transfer operations from field arithmetic so `BufferOps` and `BufferMath` are composable rather than hierarchical.
- Move Merkle tree construction out of `BufferOps` into a dedicated `Merklize` trait.
- Add buffer-native field helpers for common prover-side reductions: `ones`, `bilinear_form`, `tensor_product`, `mat_vec`, `concat`, `eq_weights`, and geometric challenge generation.
- Store IRS evaluation matrices in `Buffer` and add `values_buffer` to avoid unnecessary readbacks in prover paths.
- Update WHIR and CodeSwitch provers to use buffer-native random linear combinations and challenge groups.
## Notes For Reviewers

`geometric_challenge_groups` is intended to match drawing one host geometric challenge and slicing it into contiguous groups, but without reading the sequence back from the backend.


shreyas-londhe (approved):
LGTM

PR #267, opened 2026-07-16
Divan bench for whir

zkfriendly:
## Summary

The current Divan benchmarks only cover isolated pieces of the protocol (`sumcheck`, `expand_from_coeff`/NTT). There is no end-to-end WHIR benchmark in the Divan/CodSpeed setup. E2E WHIR timing does exist via the CLI in `src/bin/benchmark.rs`, but that tool is not a Divan bench, so it is not run in CI/CodSpeed.

This PR adds a `whir` Divan bench target with two E2E prover benchmarks on `Basefield<Field64_3>`:

- **`whir_ldt`** — WHIR as a low-degree test, with no linear forms or evaluation claims (commit + prove only).
- **`whir_pcs`** — WHIR as a polynomial commitment scheme, with 2 multilinear evaluation openings, following the PCS path in `benchmark.rs`.

Both sweep vector sizes from `2^16` to `2^21` using fixed protocol parameters (128-bit security, 20 PoW bits, folding factor 4, Blake3). Setup work (vector construction, PCS evaluation computation) stays in `with_inputs`; the timed section runs the full WHIR prover flow through a shared `run_whir` helper (domain separator, commit, prove).

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -Dwarnings`
- [x] `cargo test --release --all-features --all-targets` (includes compiling/running the new `whir` bench)
- [ ] Confirm CodSpeed picks up the new `whir` bench on merge (`cargo codspeed build` / `cargo codspeed run`)


zkfriendly:
@shreyas-londhe pending quick review here

PR #270, opened 2026-07-23
Integrate buffer abstraction to Zook

zkfriendly:
## Summary

This PR integrates the composable buffer abstraction from #266 into the current Zook, IRS, sumcheck, and code-switch protocol flow.

Once a witness is supplied as a `Buffer`, witness-sized prover state remains resident through:

`commit → IRS/NTT → sumcheck → code-switch → subsequent rounds → basecase`

This removes the two main protocol-level host bounces:

1. IRS downloading messages and masks to construct `message || mask` polynomials before the NTT.
2. Zook converting messages and covectors between `Buffer` and `Vec` on every round.

Transcript-, query-, and security-parameter-sized values intentionally remain host-side.

### Buffer-native IRS/NTT boundary

- IRS now passes resident message and mask buffers directly to the Reed–Solomon encoder.
- The CPU NTT accesses slices privately inside the CPU backend.
- Protocol code no longer constructs a witness-sized host `poly_buf`.
- Logical `message || mask` ordering and row-major `(evaluation, polynomial)` output are preserved.
- `Buffer` and `DefaultRs` are selected together through one active backend module.

### Resident Zook state

- Zook committed and round state now stores messages and covectors as `Buffer`.
- `ProtocolConfig::commit` consumes an already-resident witness buffer.
- Sumcheck outputs move directly into code-switch.
- Code-switch outputs move directly into the next round or basecase.
- The basecase-only path lifts the witness through `mixed_lift` without host materialization.

### Buffer-native code-switch arithmetic

- Code-switch messages and claim covectors are buffers.
- OOD message evaluation, covector updates, equality weights, row reductions, and in-domain dot products use buffer operations.
- Source evaluation rows are collapsed through one embedding-aware resident reduction path.
- Geometric coefficient runs are generated directly as buffers.
- OOD answers remain host-side because they are transcript-sized; their contribution is computed from the same geometric base without uploading them again.
- The original claim owns coefficient `x⁰ = 1`; OOD coefficients begin at `x¹`, and the in-domain run continues from there.

### Buffer capabilities

This PR adds or generalizes the operations required by the resident lane:

- `resize_zeroed`
- `mixed_lift`
- `mixed_mat_vec`
- `accumulate_geometric(points, scalars, prefix_len)`
- backend-native geometric challenge generation
- concatenation through `BufferOps`

The evaluator-object batching path and `UnivariateEvaluation::accumulate_many` were removed in favor of direct geometric accumulation.

### Transfer regression

A scoped regression checks that the resident IRS, Zook round, and code-switch paths do not reintroduce:

- witness-sized `to_slice()`
- `into_vec()`
- message/covector re-uploads
- OOD-answer uploads
- host `poly_buf` construction

CPU backend internals may still use slices.

## Validation

- `cargo test` — 257 passed, 30 ignored
- `cargo test --features rs_in_order` — 267 passed, 30 ignored
- `cargo clippy --all-targets -- -D warnings`
- `cargo clippy --all-targets --features rs_in_order -- -D warnings`
- resident transfer-budget regression
- `git diff --check`

## Reviewer focus

The most important invariants to review are:

1. IRS coefficient ordering remains `message_chunk || mask_row`.
2. Encoded output remains row-major `(evaluation_index, polynomial_index)`.
3. Geometric batching remains `[1, x, x², …]`.
4. ZK OOD accumulation covers the full covector while in-domain accumulation covers only `source.masked_message_length()`.
5. CPU slice access occurs only inside backend implementations, not the resident protocol path.

## Follow-ups

- [ ] Replace the transitional IRS-specific `Messages + masks` interface with one validated segmented polynomial batch. Each ordered coefficient segment should describe its resident buffers, rows per buffer, and row width. This restores the abstraction that NTT sees logical polynomials rather than message/mask semantics without physically concatenating them.

- [ ] Replace the source-text transfer regression with an instrumented `AuditBackend` that counts uploads, downloads, synchronizations, and full-buffer transfers.


zkfriendly:
@shreyas-londhe ready for review

zkfriendly:
@shreyas-londhe bumping this

src/protocols/code_switch.rs
```diff
@@ -281,35 +280,42 @@ impl<M: Embedding> Config<M> {
...
         let num_in_domain = source_evaluations.points.len();
-        let batching_coeffs =
-            geometric_challenge::<_, M::Target>(prover_state, 1 + num_ood + num_in_domain);
-        let (&original_sl_coeff, constraint_rlc_coeffs) = batching_coeffs.split_first().unwrap();
-        let (ood_rlc_coeffs, in_domain_rlc_coeffs) = constraint_rlc_coeffs.split_at(num_ood);
+        let (batching_base, batching_coeffs) = geometric_challenge_groups_with_offset::<_, M::Target>(
+            prover_state,
+            1,
+            &[num_ood, num_in_domain],
+        );
```
shreyas-londhe, line 294: prover now assumes `original_sl_coeff == 1`, but nothing says so

The verifier still applies it in both places (`original_sl_coeff * *sum`, and `covector.scalar_mul`); this side drops both. That's correct today only because `geometric_challenge` returns `geometric_sequence(F::ONE, base, count)`, so element 0 is `F::ONE` regardless of the transcript.

If the batching sequence ever starts at `x¹`, the verifier keeps reading `coeffs[0]` dynamically and stays correct while this side keeps multiplying by an implicit 1. Sums and covectors diverge, and every current test still passes because they only cover the `c₀ = 1` case.

Not asking to remove the coefficient — zook accumulates it into `round_scale_factors`. Two lines would pin the invariant instead: a note on `geometric_challenge` that index 0 is always `F::ONE`, and `debug_assert_eq!(params.original_sl_coeff, M::Target::ONE)` next to the verifier's `scalar_mul`.

src/buffer/mod.rs
```diff
@@ -166,13 +169,27 @@ pub trait BufferMath<F: Field>: Clone {
...
 
+    /// Mixed-field matrix-vector product. `self` is a row-major source-field
+    /// matrix with `vector.len()` columns; the result contains one target-field
+    /// inner product per row.
+    #[must_use]
+    fn mixed_mat_vec<M: Embedding<Source = F>>(
+        &self,
+        embedding: &M,
+        vector: &Self::TargetBuffer<M::Target>,
+    ) -> Self::TargetBuffer<M::Target>;
```
shreyas-londhe, line 187: `mixed_mat_vec` behaves exactly as `mat_vec` under the `Identity` embedding. Which renders `mat_vec` unused now. I recommend removing it.

shreyas-londhe: I now notice a similar pattern for other ops too. I suggest we remove the same-field ops if there is a mixed-field alternative, because `mixed_*` + `Identity` covers all the cases we need and we also avoid the duplication of implementations.

src/algebra/ntt/mod.rs
```diff
@@ -96,16 +126,19 @@ pub trait ReedSolomon<F>: Debug + Send + Sync {
...
     ///
     /// `codeword_length` must be NTT-smooth for this engine and at least the
     /// polynomial length.
-    fn interleaved_encode(&self, polys: &[&[F]], codeword_length: usize) -> Buffer<F>;
+    fn interleaved_encode(
+        &self,
+        messages: Messages<'_, F>,
+        masks: &Buffer<F>,
+        codeword_length: usize,
+    ) -> Buffer<F>;
```
shreyas-londhe, line 272: Pairing rules belong on the trait, not in `irs_commit`. This method now owns the `message ‖ mask` pairing that `irs_commit::commit` used to do, but the two orderings that make it well-defined aren't stated here, logical polynomial index is `vector_index * interleaving_depth + chunk_index`, and `masks[i * mask_length + c]` belongs to polynomial `i`. Both are currently documented only at `irs_commit.rs:379-382`, i.e. in one caller.

For a trait meant to be reimplemented by a GPU backend, that leaves a backend author guessing between vector-major and chunk-major flattening, with nothing asserting which they picked. Worth stating both on `interleaved_encode` (or on `Messages`).

tests/resident_transfer_budget.rs
```diff
@@ -0,0 +1,47 @@
+//! Guard the two witness-sized transfer seams fixed by the resident prover path.
+//!
+//! This deliberately checks the protocol boundary rather than CPU backend
+//! internals: CPU encoders may use slices privately, while resident protocol
+//! code must not materialize large host vectors.
```
shreyas-londhe: I agree with the thing being checked, but this is not the right way to handle this. We should use Clippy's `disallowed_methods` on `to_slice`/`into_vec` with `#[allow]` at the sites that really need a host view, opt-in instead of opt-out, enforced by CI.

src/protocols/zook/prover.rs
```diff
@@ -266,19 +257,16 @@ where
...
 
     // Prove both mask trees; subtract cs_mask contribution to project sum to f-only.
-    masker.finish(
-        &opening.round_challenges,
-        &covector[msg_len..],
-        &mut sum,
-        ps,
-    );
+    let cs_mask_indices = (msg_len..covector.len()).collect::<Vec<_>>();
+    let cs_mask_covector = covector.gather_at_indices(&cs_mask_indices);
```
shreyas-londhe: `gather_at_indices` receives an opaque index list, so a backend can't tell this is a contiguous range. On CPU that's one wasted indirection per element instead of a `memcpy`. On a device backend it's an index buffer upload plus a scatter-gather kernel to express what is one contiguous range.

shreyas-londhe: I would recommend adding a relevant method to `BufferOps` for contiguous reads.

src/protocols/zook/prover.rs
```diff
@@ -218,22 +215,16 @@ where
...
     // Sumcheck lifts the source-field message into `M::Target` at its first
-    // fold and returns the folded buffer (the covector folds in place). Move
-    // the host-side round state into buffers, fold, and move the folded result
-    // back into the `Vec` state (which downstream steps resize/truncate/index
-    // directly). The hops are zero-copy on the CPU backend.
-    let message_buf = Buffer::from(message);
-    let mut covector_buf = Buffer::from(covector);
-    let (message_buf, opening) = round.sumcheck().prove(
+    // fold. Both outputs remain resident for code-switch and the next round.
+    let mut covector = covector;
```
shreyas-londhe: Add `mut` to `covector` on L:202. So we can remove this unnecessary line.

src/algebra/ntt/cooley_tukey.rs
```diff
@@ -398,19 +402,41 @@ impl<F: Field> ReedSolomon<F> for NttEngine<F> {
...
+            .vectors
+            .iter()
+            .map(|vector| vector.to_slice())
+            .collect::<Vec<_>>();
+        let messages = vectors
+            .iter()
+            .flat_map(|vector| {
+                chunks_exact_or_empty(vector, messages.message_length, messages.interleaving_depth)
+            })
+            .collect::<Vec<_>>();
```
shreyas-londhe: This shadows the incoming argument `messages`. `polys` would be the correct naming for this as it is done below.

src/algebra/ntt/cooley_tukey.rs
```diff
@@ -398,19 +402,41 @@ impl<F: Field> ReedSolomon<F> for NttEngine<F> {
...
         }
-        let num_polys = polys.len();
-        let poly_length = polys[0].len();
-        assert!(polys.iter().all(|p| p.len() == poly_length));
+        let num_polys = messages.len();
+        let message_length = messages[0].len();
+        assert!(messages
+            .iter()
+            .all(|message| message.len() == message_length));
+        assert!(masks.len().is_multiple_of(num_polys));
```
shreyas-londhe: This will always be true because `chunks_exact_or_empty` already takes the `message_length`.

src/algebra/ntt/mod.rs
```diff
@@ -62,12 +62,42 @@ impl type_map::Family for NttFamily {
...
+    ) -> Self {
+        assert!(vectors
+            .iter()
+            .all(|vector| vector.len() == message_length * interleaving_depth));
+        Self {
+            vectors,
+            message_length,
+            interleaving_depth,
+        }
+    }
```
shreyas-londhe: All 3 fields in `Messages` are `pub`, which means that any caller can bypass the `assert` and directly create `Messages { vectors, message_length, interleaving_depth }`.  I recommend changing these to `pub(crate)`.

shreyas-londhe (changes requested):
Thanks for the PR! Overall looks correct, left some minor nits.

src/protocols/code_switch.rs
zkfriendly: > prover now assumes original_sl_coeff == 1, but nothing says so

This was already the assumption before this change: https://github.com/worldfnd/whir/blob/f9ee39741ab4c565be3a0e39fb012fbb11f07e20/src/protocols/geometric_challenge.rs#L11-L13

> If the batching sequence ever starts at x¹,

Why do we want to start batching at x¹? Particularly, it should start at 1 to support the one-constraint case without sampling unnecessary randomness in the first place.

> a note on geometric_challenge that index 0 is always F::ONE

This is [already documented](https://github.com/worldfnd/whir/blob/f92e51b7898037a86d4e795b044cf891424fd2f9/src/protocols/geometric_challenge.rs#L11-L13) too

> debug_assert_eq!(params.original_sl_coeff, M::Target::ONE)

agree with adding this

tests/resident_transfer_budget.rs
zkfriendly: ah, I did not mean to checkin this file

src/buffer/mod.rs
zkfriendly: The problem with using `Identity` + `mixed_*` everywhere is that mixed operations allocate new vectors. For example `fold` and `mixed_fold` will produce the same result under `Identity`, but `fold` will not allocate new memory and folds in place, whereas `mixed_fold` cannot do that since you need a different target buffer.

zkfriendly: The same pattern exists on the `algebra` crate, and I find it useful; for example we have `dot` which delegates to `mixed_dot`

tests/resident_transfer_budget.rs
shreyas-londhe: But this is valuable, we should have this to avoid making these buffer abstraction changes going ahead.

zkfriendly:
@BornPsych I addressed/resolved comments. The NTT interface changes are the bulk of the new diff. I originally planned the ntt interface refinement as a follow-up PR as mentioned in the description, but since we had comments here, I made the changes but we might want to iterate on the design. Next I'll update the provekit side changes

BornPsych (approved):
LGTM

shreyas-londhe (approved):

PR #271, opened 2026-08-04
Zook : Delay the witness lift to the first sumcheck fold

shreyas-londhe:
## Summary

For round-bearing plans, zook's commit currently materializes `lift(witness)` and round 0 runs its sumcheck entirely in the extension field, even though the committed data is base-field. This keeps the message in `M::Source` until the first sumcheck fold, so the largest binding runs as base x ext products and the full-size ext message allocation disappears.

### Changes

- **Buffer / algebra** : `mixed_sumcheck_polynomial` and `mixed_fold` join the existing `mixed_*` family on `BufferMath`. Both mirror `compute_sumcheck_polynomial` / `fold` exactly (including implicit zero-extension of non-power-of-two inputs) and are proptested equal to lift-then-operate over `Basefield<Field64_3>`.
- **Sumcheck** : `Config::prove` is now generic over an embedding: `a` lives in `M::Source`, crosses into the target field at the first fold, and is returned (the covector still folds in place). Same-field callers (basecase, whir, zook tail rounds) pass `Identity`, which monomorphizes to the previous code.
- **Zook** : `CommittedState::Round` and `ProverRoundState` hold the message as `Vec<M::Source>`; `prove_round` hands the round's own embedding to sumcheck (`Basefield` for round 0, `Identity` for tails). Verifier untouched.

The transcript is bit-identical to eager lifting, since the embedding is a ring homomorphism. `mixed_prove_matches_lifted_transcript` pins this down: byte-equal proofs and outputs for `prove(Basefield, message)` vs `prove(Identity, lift(message))` across arbitrary configs (Standard and ZK, masks fixed).

### Measured effect (N=2^20, `Basefield<Field64_3>`, ZK, blake3, rate 1/4, fold 3/3, 3 claims, M-series MacBook)

- Peak RSS: ~299 MB -> ~290 MB (the 24 MB ext message is replaced by the 8 MB base one; the rest of the peak is codewords and Merkle trees).
- Commit no longer allocates or writes the lifted message.
- Prove time: unchanged within noise (~0.18 s). The mixed primitives are 3-6x faster than their ext counterparts (~2.3 ms -> ~0.75 ms polynomial, ~2.8 ms -> ~0.43 ms fold at 2^20), but the whole round-0 message side is only ~5 ms of the prove, so this is a memory and structure win, not a latency one.

## Notes For Reviewers

- `prove`'s signature change is the reason basecase and whir call sites appear in the diff; those edits are mechanical (`Identity`, rebind the returned buffer).
- With `num_rounds == 0` the sumcheck performs no fold, so `prove` falls back to a plain lift to keep returning a target-field buffer. Zook only hits this through basecase-only plans, which lift at commit as before.
- The pre-fold source buffer is dropped without wiping. The eager path had the same exposure class (in-place `fold` frees the truncated half via `shrink_to_fit`), so this does not regress hiding, but a `wipe()` could be added if we want stricter prover memory hygiene.
- Cross-branch check (not in-repo): a fixed-seed Standard-mode roundtrip hashes to the same proof bytes on main and on this branch. ZK proofs are run-to-run nondeterministic (mask sampling uses the entropy-seeded prover RNG), which is why the in-repo equivalence test fixes the masks instead.

## Test plan

- `cargo test --release` : full suite green (257 passed), including the Basefield roundtrips that now exercise the mixed path against the unchanged verifier.
- New tests: `mixed_sumcheck_poly_matches_lifted`, `mixed_fold_matches_lifted` (algebra-level proptests, lengths 0..2^10 incl. non-powers-of-two), `mixed_prove_matches_lifted_transcript` (transcript identity, arbitrary configs).
- `cargo clippy --all-targets` clean.

### DIFF
--- a/src/protocols/sumcheck.rs
+++ b/src/protocols/sumcheck.rs
@@ -527,6 +527,63 @@ mod tests {
         });
     }
 
+    /// The delayed lift's core claim: proving with a source-field `a` through
+    /// a real embedding is byte-identical (proof and outputs) to lifting `a`
+    /// up front and proving through `Identity`.
+    #[test]
+    fn mixed_prove_matches_lifted_transcript() {
+        use crate::algebra::{embedding::Basefield, fields::Field64_3, lift};
+        crate::tests::init();
+        let embedding = Basefield::<Field64_3>::new();
+        proptest!(|(seed: u64, config in Config::<Field64_3>::arbitrary())| {
+            let instance = U64(seed);
+            let ds = DomainSeparator::protocol(&config)
+                .session(&format!("Mixed vs lifted at {}:{}", file!(), line!()))
+                .instance(&instance);
+            let mut rng = StdRng::seed_from_u64(seed);
+            let a_source: Vec<Field64> = random_vector(&mut rng, config.initial_size);
+            let covector: Vec<Field64_3> = random_vector(&mut rng, config.initial_size);
+            let masks: Vec<Field64_3> =
+                random_vector(&mut rng, config.mask_length() * config.num_rounds);
+            let a_lifted = lift(&embedding, &a_source);
+            let initial_sum = dot(&a_lifted, &covector);
+
+            let run = |mixed: bool| {
+                let mut b = Buffer::from(covector.as_slice());
+                let mut sum = initial_sum;
+                let mut prover_state = ProverState::new_std(&ds);
+                let (folded, opening) = if mixed {
+                    config.prove(
+                        &mut prover_state,
+                        &embedding,
+                        &Buffer::from(a_source.as_slice()),
+                        &mut b,
+                        &mut sum,
+                        &masks,
+                    )
+                } else {
+                    config.prove(
+                        &mut prover_state,
+                        &Identity::new(),
+                        &Buffer::from(a_lifted.as_slice()),
+                        &mut b,
+                        &mut sum,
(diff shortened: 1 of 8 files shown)


src/protocols/sumcheck.rs
```diff
@@ -175,14 +184,25 @@ impl<F: Field> Config<F> {
...
         {
             // Fold and compute sumcheck polynomial in one pass.
-            let (c0, c2) = if let Some(w) = prev_round_challenge {
-                a.fold_pair_sumcheck_polynomial(b, w)
+            let (c0, c2) = if let Some(folded) = a_folded.as_mut() {
+                let w = prev_round_challenge.expect("folded buffer implies a prior challenge");
+                folded.fold_pair_sumcheck_polynomial(b, w)
+            } else if let Some(w) = prev_round_challenge {
+                // First fold: `a` crosses from the source to the target field.
+                let folded = a.mixed_fold(embedding, w);
```
zkfriendly: Just noting that when the embedding is Identity, the original full-size buffer stays alive while this creates a new half-size folded buffer. So during the fold we temporarily hold about 1.5× the message data, whereas this was previously folded in place. I don't think we can avoid this for base->ext but for Identity we can.

zkfriendly (approved):
LGTM! Left a minor comment.

PR #274, opened 2026-08-20
Zook: return the final evaluation point from `prove`

BornPsych:
Stacked on #270 — review that first; this PR's diff is only the last commit.

## What

`ProtocolConfig::prove` currently returns `()` and discards everything the transcript derived, so a caller that wants to continue reasoning about the proof (e.g. wiring Zook into an outer protocol, or checking prover/verifier agreement in tests) has no way to get at the final evaluation point. This PR returns it.

`prove` now returns a `ProverClaim<F>`:

```rust
pub struct ProverClaim<F: ark_ff::Field> {
    /// Sumcheck challenges followed by the basecase evaluation points.
    pub evaluation_point: Vec<F>,
    /// Fiat-Shamir RLC coefficients for the input forms.
    pub rlc_coefficients: Vec<F>,
}
```

These are the prover-side mirror of the `FinalClaim` fields the verifier already returns.

## How

- `ProverRoundState` gains `all_round_challenges: Vec<M::Target>`, accumulating each round's sumcheck challenges in transcript order.
- After the rounds, the basecase opening's `evaluation_points` are appended, giving the full point in the same order the verifier reconstructs it.
- A `debug_assert` pins `evaluation_point.len() == log2(vector_size)`.
- The basecase opening was previously bound to `let _ =`; it is now used.

## Tests

- `build_and_prove` returns a `Proven` struct instead of a 4-tuple — the tuple was already 4 fields wide and this adds a 5th.
- Roundtrip tests assert the prover's `evaluation_point` and `rlc_coefficients` match the verifier's `FinalClaim`, so a divergence between the two sides fails loudly rather than silently.

`cargo test zook` → 21 passed. `cargo fmt --check` and `cargo clippy --all-targets` clean.

### DIFF
--- a/src/protocols/zook/mod.rs
+++ b/src/protocols/zook/mod.rs
@@ -59,11 +59,20 @@ impl<F: ark_ff::Field> FinalClaim<F> {
     }
 }
 
+/// Values derived by [`ProtocolConfig::prove`] from the prover transcript.
+#[derive(Clone, Debug)]
+pub struct ProverClaim<F: ark_ff::Field> {
+    /// Sumcheck challenges followed by the basecase evaluation points.
+    pub evaluation_point: Vec<F>,
+    /// Fiat-Shamir RLC coefficients for the input forms.
+    pub rlc_coefficients: Vec<F>,
+}
+
 #[cfg(test)]
 mod tests {
     use ark_std::rand::{rngs::StdRng, SeedableRng};
 
-    use super::ProtocolConfig;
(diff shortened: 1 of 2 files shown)


zkfriendly (changes requested):
Looks great! thanks. Only thing is CI is failing due to a Clippy error; it's coming from an unchanged file, but it would be nice to fix it here and then merge with a green CI.

zkfriendly (approved):

